PR #2, opened 2016-07-12
Added tests!

czardoz:
Should merge #1 before this.

### DIFF
--- a/lib/runner/extensions/event.command.js
+++ b/lib/runner/extensions/event.command.js
@@ -93,7 +93,7 @@ module.exports = {
                     }),
                 stopOnScriptError = _.has(payload, 'stopOnScriptError') ? payload.stopOnScriptError :
                     this.options.stopOnScriptError,
-                abortOnError = this.options.abortOnError,
+                abortOnError = _.has(payload, 'abortOnError') ? payload.abortOnError : this.options.abortOnError,
                 events;
 
             // validate the payload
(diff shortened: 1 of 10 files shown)


lib/runner/extensions/event.command.js
```diff
@@ -161,7 +162,7 @@ module.exports = {
             }, function (err, results) {
                 // trigger the event completion callback
                 util.safeCall(this.triggers[eventName], this.triggers, null, cursor, results, item);
-                next(null, results, err);
```
czardoz: This corrects the existing behavior which treats a request error as fatal.

lib/runner/extensions/event.command.js
```diff
@@ -93,6 +93,7 @@ module.exports = {
                     }),
                 stopOnScriptError = _.has(payload, 'stopOnScriptError') ? payload.stopOnScriptError :
                     this.options.stopOnScriptError,
+                abortOnError = this.options.abortOnError,
```
shamasis: Why also not take this from payload like stopOnError does?

lib/runner/extensions/event.command.js
```diff
@@ -151,7 +152,7 @@ module.exports = {
                     this.triggers.script(err || null, cursor, result, script, event, item);
 
                     // move to next script and pass on the results for accumulation
-                    next((stopOnScriptError && err) ? err : null, _.assign({
+                    next(((stopOnScriptError || this.options.abortOnError) && err) ? err : null, _.assign({
```
shamasis: Why use `this.options.abortOnError` when `abortOnError` is defined on top?

lib/runner/extensions/event.command.js
shamasis: Rewrite with `(abortOnError && err) ? err : null`

PR #3, opened 2016-07-12
Added ability to gracefully stop an iteration when there's an error

czardoz:
Implemented the `stopOnError` option.

Should be merged after #2


lib/runner/extensions/request.command.js
```diff
@@ -59,7 +59,7 @@ module.exports = {
                         item.request, payload.item,
                     // todo: remove these asap
                     legacyResponse, legacyRequest);
-                next(err, response, item.request, legacyResponse, legacyRequest);
+                next((err && this.options.abortOnError) ? err : null, response, item.request, legacyResponse, legacyRequest);
```
shamasis: Is this not defined as `abortOnError` on top?

PR #4, opened 2016-07-12
implemented the stop-on-runner flag

czardoz:
We now stop the run gracefully

### DIFF
--- a/lib/runner/extensions/item.command.js
+++ b/lib/runner/extensions/item.command.js
@@ -16,12 +16,13 @@ module.exports = {
 
     process: {
         /**
+         * @param {Function=} callback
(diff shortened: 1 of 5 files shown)


lib/runner/extensions/item.command.js
```diff
@@ -91,7 +92,15 @@ module.exports = {
                     environment: ctxTemplate.environment,
                     data: ctxTemplate.data,
                     coords: coords
-                }).done(function (response, request, legacyResponse, legacyRequest) {
+                }).done(function (response, request, legacyResponse, legacyRequest, requestError) {
```
shamasis: lets make this object?

PR #11, opened 2016-07-20
allow setting duplicate headers in requests

czardoz:
Also added an integration test


lib/requester/util.js
```diff
@@ -45,7 +46,7 @@ module.exports = {
         if (bodyParams && bodyParams.body) {
             self.ensureHeaderExists(options.headers, 'Content-Type', 'text/plain');
         }
-        self.ensureHeaderExists(options.headers, 'User-Agent', 'PostmanRuntime/0.0.1');
```
abhijitkane: tsk
molecular commit

czardoz: ('-,-)

PR #12, opened 2016-07-21
Don't send empty request bodies

czardoz:
Currently, if the request body is empty, runtime still adds a header for form-data (the form-data boundary).

This PR fixes this.


czardoz:
Also, tests have been added:

PR #13, opened 2016-07-22
handle files more intelligently in the requester

czardoz:
This change ensures that the requester does not try to send any file data if no file resolver is provided to it.
Also, added tests for this.


lib/requester/util.js
```diff
@@ -146,8 +146,15 @@ module.exports = {
...
-                        param.value = defaults.fileResolver ? defaults.fileResolver.createReadStream(param.src) : '';
+                    var isFile = (param.type === 'file'),
+                        fileResolver = defaults.fileResolver;
+
+                    if (isFile && !fileResolver) { // It's a file, but we don't have a resolver for it
+                        return accumulator;
+                    }
+
+                    if (isFile && param.src) {
+                        param.value = fileResolver ? fileResolver.createReadStream(param.src) : '';
```
elssar: You don't need to check whether `fileResolver` exists again

lib/runner/extensions/request.command.js
```diff
@@ -15,6 +52,8 @@ module.exports = {
         // @todo - remove this when requester creation is offloaded to runner
         if (this.options.requester && this.options.requester.external === true) {
             this.requester = this.options.requester.instance;
+            // Just so we don't break the App, pull the file resolver out of the requester
```
shamasis: How do we set file resolver where we are not setting requester externally?

shamasis:
LGTM

PR #16, opened 2016-07-28
ensure XHR compatible url-encoding of parameters

czardoz:
Ensures that "request" creates URL encoded formdata or querystring as
```
foo=bar&foo=baz
``` 
instead of 
```
foo[0]=bar&foo[1]=baz
```


elssar:
LGTM :+1:

shamasis:
what was it affecting?

czardoz:
@shamasis It was a behavioral change for the apps.

PR #25, opened 2016-08-01
Added content-encoding handlers

vegetableman:
Experimental. But gets the correct size for encoded content `gzip` and `deflate` for both browser and node.

### DIFF
--- a/lib/requester/requester.js
+++ b/lib/requester/requester.js
@@ -65,7 +65,9 @@ Requester.prototype.request = function (item, cb, scope) {
             strictSSL: this.strictSSL,
             followRedirects: this.followRedirects
         }),
-        startTime = Date.now();
+        startTime = Date.now(),
+        rawResponse = '';
+
 
     return requests(requestOptions, function (err, res, resBody) {
         if (err) {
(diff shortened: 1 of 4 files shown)


lib/requester/requester.js
```diff
@@ -99,17 +101,34 @@ Requester.prototype.request = function (item, cb, scope) {
...
+    })
+    .on('response', function(response) {
+        if(response.headers['content-encoding']) {
+            var chunks = [];
+            response.on('data', function(chunk) {
+                chunks.push(chunk);
+            })
+            .on('end', function() {
+                chunks.forEach(function(chunk) {
+                    rawResponse += util.arrayBufferToString(chunk);
```
shamasis: This is very costly. Use same buffer byte length.

package.json
```diff
@@ -30,6 +30,7 @@
     "crypto-js": "3.1.6",
     "lodash": "3.10.1",
     "node-uuid": "1.4.7",
+    "pako": "1.0.3",
```
czardoz: Why not use node's Zlib?

czardoz: If we want things to work in the browser too, we can add a browserify/webpack rule

vegetableman: for node, `request` module itself flushes out gzipped (raw response) content on `response` event. (see above).  
`pako` is only used for gzipping browser response body.

lib/requester/requester.js
```diff
@@ -99,17 +101,34 @@ Requester.prototype.request = function (item, cb, scope) {
...
                 cookie: undefined, // @todo get from cookieJar and implement loading/unloading in SDK
-                responseTime: responseTime
+                responseTime: responseTime,
+                encodedData: encodedData
             };
 
         cb.call(scope || this, null, response, actualRequest, responseObj, request);
-    });
+    })
+    .on('response', function(response) {
```
czardoz: This is not compatible with the browsers, so you cannot use this event

lib/requester/requester.js
vegetableman: I am returning the rawResponse instead of calculating here, so that it could be displayed on the app later, calculation could be done on the `response` in `postman-collection`. Also almost all the api-related tools, display the encoded content for gzipped content.

lib/requester/requester.js
vegetableman: Added a `on` function to the ` xhr` object to ensure compatibility.

lib/requester/requester.js
vegetableman: Have updated the code to make a single call to `arrayBufferToString`

shamasis: Even then. Currently app is slow on large response only because of this reason - that we convert to string right away. Is it good to add one more thing that adds to this burden? Ideally we could expose only the raw size and later expose raw response buffer itself so that it can be consumed directly from the buffer in app and Newman. 

@czardoz thoughts?

package.json
shamasis: Means that we are re-encoding it to get the gzipped body back?bits giving me the heebee geebeez 😶

vegetableman:
Simplified it to include only the raw buffer. All the size related calculation could be moved to the response `size` function in `postman-collection`. Will that do?

lib/requester/requester.js
```diff
@@ -105,11 +107,23 @@ Requester.prototype.request = function (item, cb, scope) {
...
+        if(response.headers['content-encoding']) {
+            var chunks = [];
+            response.on('data', function(chunk) {
+                chunks.push(chunk);
+            })
+            .on('end', function() {
+                rawResponse = Buffer.concat(chunks);
```
shamasis: This is brillian.

shamasis:
Closing this since we intend to re work and revise this when the requester interface is converted to a full event model.

PR #30, opened 2016-08-04
allow client-side ssl auth in the requester

czardoz:
Initial work on Client side certificates


lib/requester/requester.js
```diff
@@ -45,6 +46,7 @@ var requests = require('./request-wrapper'),
         this.timeout = _.has(options, 'timeout') ? options.timeout : undefined;
         this.fileResolver = _.get(options, 'fileResolver');
         this.strictSSL = _.get(options, 'strictSSL');
+        this.certificateManager = options.certificateManager;
```
shamasis: _.get like file resolver?

shamasis:
Test fails.

PR #31, opened 2016-08-05
Two changes - binary file upload and disabled header ignore

czardoz:
* Added support for binary file uploads
* Ensured that disabled headers are not sent


lib/requester/util.js
```diff
@@ -42,6 +42,7 @@ module.exports = {
         options.followAllRedirects = defaultOpts.followRedirects;
 
         // Request body may return different options depending on the type of the body.
+        console.log(request);
```
shamasis: WHAAAT?

lib/requester/util.js
```diff
@@ -88,6 +89,8 @@ module.exports = {
         var headers = {};
         if (request.headers) {
             request.forEachHeader(function (header) {
+                if (header.disabled) { return; }
```
shamasis: if auth or other stuff adds headers that were disabled then?

lib/requester/util.js
```diff
@@ -165,6 +168,11 @@ module.exports = {
                 }, {})
             };
         }
+        else if (request.body.mode === 'file') {
+            options = {
+                body: _.get(request, 'body.file') ? _.get(request, 'body.file').content : undefined
```
shamasis: _.get(request, 'body.file.content')

lib/requester/util.js
shamasis: https://github.com/postmanlabs/postman-collection/pull/81 fixes it

lib/runner/extensions/request.command.js
```diff
@@ -13,31 +13,41 @@ var _ = require('lodash'),
      * @param callback
      */
     resolveFiles = function (request, resolver, triggers, callback) {
-        var formParams = _.get(request, 'body.formdata');
+        var formParams = _.get(request, 'body.formdata'),
+            fileData = _.get(request, 'body.file');
 
-        if (!formParams) { return callback(null, request); }
+        if (!(formParams || fileData)) { return callback(null, request); }
```
shamasis: do stuff based on mode, else two callbacks would be called + if binary data is there, but form data is used in mode, both will get uploaded.

PR #35, opened 2016-08-10
ensure that case-insensitive cookies are actually case insensitive

czardoz:


abhijitkane:
Uncaught Exception: Cannot call toLowerCase() of undefined

lib/uvm/browser/sandbox-bus-start.js
```diff
@@ -236,7 +236,7 @@
 
         if (options.responseCookies) {
             for (i = 0; i < options.responseCookies.length; i++ ) {
-                insensitiveCookies[options.responseCookies[i].name] = options.responseCookies[i];
+                insensitiveCookies[options.responseCookies[i].name.toLowerCase()] = options.responseCookies[i];
```
elssar: Check that `name` exists and is a string

PR #42, opened 2016-08-12
added a test to check for sent http headers being bubbled up

czardoz:


czardoz:
Merging since it's only a test

PR #45, opened 2016-08-12
stringify the url before calling the proxy manager

czardoz:


czardoz:
Merging with @akshay's approval

PR #51, opened 2016-08-24
ensure that we parse the url again after resolving variables

czardoz:
which is a solution for:
* postmanlabs/postman-app-support#2276
* postmanlabs/postman-app-support#2285


shamasis:
:shipit:

PR #57, opened 2016-09-02
Added appveyor build configuration

kunagpal:


shamasis:
Great!

Can you add a tiny system test to ensure that node versions mentioned in travis matches appveyor. You'll also see standard system tests for travis YML that you can add for appveyor file.

kunagpal:
Good idea. I'll add them to newman as well.

PR #59, opened 2016-09-07
Fixed a memory issue

czardoz:
Running SugarJS with each execution causes leaks!
Now, we only run it once for each host.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,38 @@
+#---------------------------------#
+#    environment configuration    #
+#---------------------------------#
+
+# scripts that are called at very beginning, before repo cloning
+init:
+  # So that the build output does not cause windows to fail.
+  - git config --global core.autocrlf input
+
+# environment variables
+environment:
+  matrix:
+    - nodejs_version: "4"
+    - nodejs_version: "6"
+
+# scripts that run after cloning repository
+install:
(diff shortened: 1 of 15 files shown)


lib/uvm/vm-require.js
```diff
@@ -1,20 +1,19 @@
...
+    var filePath = require.resolve(moduleId),
+        tmp = require.cache[filePath],
         ret;
-    require.cache = _cache;
-    ret = require(moduleId);
-    require.cache = tmp;
+
+    delete require.cache[filePath];
+    ret = require(filePath);
+    require.cache[filePath] = tmp;
```
shamasis: delete and restore only if previously cached.

shamasis:
For some reasons all checks are failing.

appveyor.yml
```diff
@@ -4,13 +4,13 @@
...
 # scripts that are called at very beginning, before repo cloning
 init:
+  # So that the build output does not cause windows to fail.
   - git config --global core.autocrlf input
 
 # environment variables
 environment:
   matrix:
     - nodejs_version: "4"
-    - nodejs_version: "5"
```
shamasis: Why?

shamasis:
Merged since travis PR status passed

PR #65, opened 2016-09-12
Replaced http-codes.js with http-reasons

kunagpal:


shamasis:
Module usage does not match and tests needed to address current failing scenario

kunagpal:
Yes, I realized that. Will be adding tests in the next commit.

czardoz:
@kunagpal Was this change even tested out? The `http-reasons` module exports a `lookup` function. It's not an object you can just read properties from. :confused:

test/integration-legacy/runner-spec.test.js
```diff
@@ -250,6 +251,7 @@ describe('Runner', function () {
                             expect(cursor.ref).to.eql(runStore.ref);
 
                             expect(response.code).to.be(200);
+                            expect(reason.lookup(response.code).name).to.be(response.status);
```
czardoz: This should be the other way round. Actual things should come in the `expect` call, and expected results compared with them.

kunagpal: Okay, this condition will be reversed in the next commit.

test/integration-legacy/runner-spec.test.js
```diff
@@ -251,7 +251,7 @@ describe('Runner', function () {
                             expect(cursor.ref).to.eql(runStore.ref);
 
                             expect(response.code).to.be(200);
-                            expect(reason.lookup(response.code).name).to.be(response.status);
+                            expect(response.status).to.be(reason.lookup(response.code).name);
```
shamasis: This is not a valid test. You're looking up a system that itself is generating the assertion!!!

Match the status and detail with manual text.

kunagpal: Wouldn't that end up being a test for `http-reasons`?

kunagpal: Done

PR #66, opened 2016-09-14
Added moment.js to the sandbox

kunagpal:
TODO:
- [ ] Write tests
- [ ] Add moment in browser sandbox


shamasis:
@kunagpal - resend with tests and inclusion in browser sandbox.

PR #75, opened 2016-09-22
ensure that 'false' is casted to a string when using postman api

czardoz:
A flawed check for value caused the value to not be set as a string if it is `false`. That is now corrected.


czardoz:
@kunagpal Good catch. However, these parts of the code are executed at runtime, in a sandboxed environment or an iframe. We do not exactly have access to `require`s here.

kunagpal:
@czardoz In that case, LGTM.

PR #82, opened 2016-09-27
ensure that user-set globals are available in the sandbox

czardoz:


lib/uvm/browser/sandbox-bus-start.js
```diff
@@ -4,10 +4,17 @@
         Scope;
 
     Scope = function (replacements) {
-        var globals = Object.keys(win),
+        var globals,
             object = {},
             transferables = [];
 
+        if (!Array.isArray(win.__js_globals)) {
+            win.__js_globals = Object.keys(win);
```
shamasis: don't inject this back

lib/uvm/browser/sandbox-bus-start.js
```diff
@@ -74,27 +81,6 @@
         'eval'
     ];
 
-    Scope.trackedGlobals = Object.keys(win);
```
shamasis: dont delete. comment

PR #84, opened 2016-09-27
ensure that we don't override content type if one is already given

czardoz:
This fixes a bug which causes duplicate content type to be added sometimes.

### DIFF
--- a/lib/requester/browser/request.js
+++ b/lib/requester/browser/request.js
@@ -128,7 +128,11 @@ function request(options, callback) {
         }
         if((new RegExp(METHODS_WITH_BODY.join('|'), 'i')).test(options.method)){
             var encoding = (options.encoding || 'application/x-www-form-urlencoded').toLowerCase();
-            options.headers['content-type'] = encoding;
+
+            if (!options.headers['content-type'] && !options.headers['Content-Type']) {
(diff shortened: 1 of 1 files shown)


lib/requester/browser/request.js
```diff
@@ -128,7 +128,11 @@ function request(options, callback) {
         }
         if((new RegExp(METHODS_WITH_BODY.join('|'), 'i')).test(options.method)){
             var encoding = (options.encoding || 'application/x-www-form-urlencoded').toLowerCase();
-            options.headers['content-type'] = encoding;
+
+            if (!options.headers['content-type'] && !options.headers['Content-Type']) {
+                options.headers['content-type'] = encoding;
+            }
+
```
vegetableman: It could be made stricter similar to this:-
https://github.com/request/request/blob/bfb3a469a1c51ebf6a3c1f4be553b43eaafcf47c/request.js#L1137

czardoz: That replaces the header if it exists. This does not. In fact, we should update `postman-request` to behave similarly to this.

vegetableman: Looks Good. Do note: This fix also needs to be present on `Postman-request` to ensure consistency.

PR #88, opened 2016-09-30
remove everything from the global scope

czardoz:
Remove all injected objects from globals, and instead use variable references to them.

### DIFF
--- a/lib/uvm/vm-execute.js
+++ b/lib/uvm/vm-execute.js
@@ -7,35 +7,51 @@
         postmanApi,
         scope,
 
-        replacements = {
-            setTimeout: function (fn, ms) {
-                return setTimeout(function () {
-                    try { return fn.apply(this, arguments); }
-                    catch (e) {
-                        universe.on.exception(e);
-                    }
-                }, ms);
-            },
+        // references for us to use
+        on = universe.on,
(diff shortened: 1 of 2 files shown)


lib/uvm/vm-execute.js
```diff
@@ -7,35 +7,51 @@
...
+        __done = universe.__done,
+        __async = universe.__async,
+        __delete = universe.__delete,
+        __require = universe.__require,
+        __masked = universe.__masked,
+        requires = universe.requires,
+        errors = universe.errors,
+        replacements;
+
+    __delete && __delete.length && __delete.forEach(function(key) {
```
czardoz: Who will delete `__delete`! :D

shamasis: lol!

lib/uvm/vm-execute.js
```diff
@@ -46,7 +62,7 @@
         }
         catch (e) {
             libraries[lib.object] = undefined;
-            universe.on.exception(e); // We'll try to run without this library
+            on.exception(e); // We'll try to run without this library
```
shamasis: where is `on` defined?

shamasis: found it

PR #111, opened 2016-10-13
* Ensure that the response reason phrase is used if available

czardoz:
* cleanup of the requester code

### DIFF
--- a/lib/requester/requester.js
+++ b/lib/requester/requester.js
@@ -79,42 +79,66 @@ Requester.prototype.request = function (item, cb, scope) {
             return cb.call(scope || this, err);
         }
 
-        var responseString = (resBody != null && resBody.toString) ? resBody.toString() : resBody;
+        var responseString,
+            responseTime,
+            responseJSON,
+            legacyRequest,
+            legacyResponse,
+            cookies,
+            codeReason,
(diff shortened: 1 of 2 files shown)


lib/requester/requester.js
```diff
@@ -79,42 +79,68 @@ Requester.prototype.request = function (item, cb, scope) {
...
+
+        // Pull out cookies from the cookie jar, and make them chrome compatible.
+        cookies = (cookieJar && _.isFunction(cookieJar.getCookies)) ?
+            _.transform(cookieJar.getCookies(requestOptions.url), function (acc, cookie) {
+                acc.push(toChromeCookie(cookie));
+            }, []) : [];
+
+        // get the reason phrase
+        codeReason = httpReasons.lookup(responseJSON.statusCode);
+        codeReason.overridden = !(res && res.statusMessage);
```
shamasis: `res && res.statusMessage` is repeated

```
codeReason = httpReasons.lookup(responseJSON.statusCode);

if (res && res.statusMessage && (codeReason.name !== res.statusMessage)) {
    codeReason.name = res.statusMessage;
    codeReason.nonstandard = true;
}
```

something like above

PR #122, opened 2016-10-18
Improved .ignore system tests

kunagpal:


kunagpal:
@czardoz @shamasis Mergable?

PR #150, opened 2016-11-16
Replaced Lodash with Lodash3 in sandbox

kunagpal:
Also added a system test to ensure `lodash3` sanity.

### DIFF
--- a/lib/uvm/vm-execute.js
+++ b/lib/uvm/vm-execute.js
@@ -20,6 +20,7 @@
         Scope = universe.Scope,
         PostmanApi = universe.PostmanApi,
         code = universe.code,
+        required,
         replacements;
 
     __delete && __delete.length && __delete.forEach(function(key) {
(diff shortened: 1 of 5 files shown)


lib/uvm/vm-execute.js
```diff
@@ -61,7 +62,8 @@
         lib = requires[i];
         // Play around with the require cache
         try {
-            libraries[lib.object] = __require(lib.require);
+            required = __require(lib.require);
+            libraries[lib.object] = lib.require === 'lodash3' ? required.noConflict() : required;
```
czardoz: instead of adding a string like this and creating another special case, why not check if noConflict exists and is a function?

kunagpal: Makes sense, but what if .noConflict exists for other libraries as well? Do we want to require without conflicts there as well?

shamasis: I agree with Kunal here. Since we do not have pre and post processors, it is only way that we create debt now and later recover with our browserify changes. Possibly add this as a todo note here before merging.

czardoz: Good idea

kunagpal:
@czardoz @shamasis The todo has been added.

PR #154, opened 2016-11-19
Improved legacy integration tests

kunagpal:
* Prevents multi callback errors from masking the actual test errors.
* Saves repetitive debugging effort.


shamasis:
The overall approach seems okay. Since the legacy test itself are a bit in technical debt, I am not deep diving into any improvements. It's good to go from my end.

@czardoz wrote the original set, his input would be more important.

kunagpal:
@shamasis: @czardoz and I have indeed discussed this at length.

PR #164, opened 2016-12-09
ensure that the final callback is not called more than once

czardoz:


lib/runner/run.js
```diff
@@ -67,10 +67,13 @@ _.extend(Run.prototype, {
             return callback(new Error('run: already running'));
         }
 
-        _.isFinite(_.get(this.options, 'timeout.global')) &&
-            backpack.timeback(callback, this.options.timeout.global, this, function () {
```
shamasis: Why not just do `(callback = …)` here?

czardoz: `callback` is later used as the `run.triggers` object.

czardoz: I mean, `backpack.timeback` removes the normalized functions from the `callback`, and that is later used for the `run.triggers`.

PR #169, opened 2016-12-27
Removed the existing proxymanager integration and added the proxyList integration for Custom proxy support

SamvelRaja:


czardoz:
@shamasis Can you review this breaking change? We can release this with v4.0

SamvelRaja:
@kunagpal you need to update newman also. If **proxyManager** is already implemented in newman.

shamasis:
Checking.

lib/requester/request-wrapper.js
```diff
@@ -78,16 +79,18 @@ var _ = require('lodash'),
...
-        if (url && options.proxyManager && options.proxyManager.getProxyConfiguration) {
+        if (url && options.proxyList) {
+            if(!_.isFunction(options.proxyList.isProxyConfigList)) {
+              options.proxyList = new ProxyConfigList({}, options.proxyList);
+            }
             _.isFunction(url.toString) && (url = url.toString());
-            return options.proxyManager.getProxyConfiguration(url, function(err, obj) {
-                options.proxy = obj && obj.proxy;
-                options.tunnel = obj && obj.tunnel;
-                _.set(options, 'request._._postman_proxy', obj);
```
SamvelRaja: @czardoz 
`_.set(options, 'request._._postman_proxy', obj);` I removed this by mistake, may i know why this is for?

lib/requester/request-wrapper.js
```diff
@@ -78,16 +79,18 @@ var _ = require('lodash'),
     setProxy = function(options, cb) {
         var url = _.get(options, 'request.url');
 
-        if (url && options.proxyManager && options.proxyManager.getProxyConfiguration) {
+        if (url && options.proxyList) {
+            if(!_.isFunction(options.proxyList.isProxyConfigList)) {
+              options.proxyList = new ProxyConfigList({}, options.proxyList);
```
shamasis: Runtime assumes that the incoming parameters are sanitised. Can we not assume that incoming proxyList option if truthy, is a proxy list and proceed without cloning it?

That would also mean no dependency on collection SDK. Also it would not clone the list and use existing object. It is the consumer of runtime who decides what to give cloned and runtime can mutate the options - like it does for variable scope.

lib/requester/request-wrapper.js
vjrngn: From the context of the app, this isn't required.

czardoz: @vjrngn @SamvelRaja This is used inside the app Console to log the detected proxy version.

vjrngn: Oh I see. Didn't know.

lib/requester/request-wrapper.js
SamvelRaja: 👍

czardoz: @shamasis `options.proxyList` can be an instance of `ProxyList` or a POJO. Since the list is created in the browser process, it cannot be passed as-is into the Node process. Which is why this is correct IMO.

test/integration/sanity/proxy-http.test.js
```diff
@@ -1,13 +1,18 @@
...
+        _ = require('lodash'),
         proxy = require('http-proxy'),
 
         server,
         testrun;
 
     before(function (done) {
         var port = 9090,
-            proxyList = [{match: '*://*.getpostman.com/*', server: 'http://localhost:' + port, tunnel: false}];
+            proxyList = new ProxyConfigList({}, [{
```
SamvelRaja: @vjrngn you need to pass the proxyList as sdk Instance instead of a normal ListArray.

lib/requester/request-wrapper.js
```diff
@@ -88,6 +88,7 @@ var _ = require('lodash'),
             if(proxyConfig) {
               options.proxy = proxyConfig.server.toString(),
               options.tunnel = proxyConfig.tunnel;
+              _.set(options, 'request._._postman_proxy', proxyConfig);
```
SamvelRaja: Also, @vjrngn  the meta schema has been changed from
{```

  proxy: 'http://proxy.com/
 tunnel: true
}
```
to 
```
{
 server: 'http://proxy.com/
 tunnel: true
}
```

SamvelRaja:
@shamasis Updated the changes

PR #171, opened 2017-01-06
Updated postman-sandbox to v1.0.1 which fixes issue with runtime not initialising in early Node v4.x

shamasis:


vjrngn:
Looks good.

PR #172, opened 2017-01-10
Added integration tests that fetch latest newman and run test with runtime replaced in it.

shamasis:
This PR adds a new test command: `npm run test-integration-newman`.

This is not executed as part of npm test and instead it is meant to be executed on demand

### DIFF
--- a/npm/test-integration-newman.js
+++ b/npm/test-integration-newman.js
@@ -1,0 +1,65 @@
+#!/usr/bin/env node
+require('shelljs/global');
+require('colors');
+
+var path = require('path'),
+    async = require('async'),
+    tmp = require('tmp');
+
+module.exports = function (exit) {
+    // banner line
+    console.log('Running newman integration tests...'.yellow.bold);
+
+    tmp.dir(function (err, dir, cleanup) {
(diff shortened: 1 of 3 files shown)


npm/test-integration-newman.js
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env node
+require('shelljs/global');
+require('colors');
+
+var path = require('path'),
+    async = require('async'),
+    tmp = require('tmp');
```
kunagpal: Consider using the native `dir = require('os').tmpdir()` to get a temporary directory, instead of adding a new devDependency.

npm/test-integration-newman.js
```diff
@@ -0,0 +1,65 @@
...
+    tmp.dir(function (err, dir, cleanup) {
+        if (err || dir.length < 4) {
+            console.error(err);
+            return exit(1);
+        }
+
+        var installDir = path.join('node_modules', 'newman');
+
+        pushd(dir);
+        async.waterfall([
```
kunagpal: `async.series`

PR #184, opened 2017-03-07
Fix issue in setting certificate file contents

kamalaknn:
1. Fix typo in error handling
  2. Fix usage of `async.mapValues`
  3. Fix tests to fail if certificate is invalid

### DIFF
--- a/lib/requester/request-wrapper.js
+++ b/lib/requester/request-wrapper.js
@@ -115,11 +115,11 @@ var _ = require('lodash'),
         }
 
         if (!_.isString(certPath)) {
-            returncb(new Error(ERROR_MISSING_CERT_PATH));
+            return cb(new Error(ERROR_MISSING_CERT_PATH));
         }
 
-        async.mapValues([keyPath, certPath], function (value, key, cb) {
-          resolver.readFile(value, cb);
+        async.mapValues({key: keyPath, cert: certPath}, function (value, key, callback) {
+            resolver.readFile(value, callback);
         }, function (err, fileContents) {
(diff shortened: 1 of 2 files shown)


lib/requester/request-wrapper.js
```diff
@@ -115,21 +115,19 @@ var _ = require('lodash'),
...
 
         if (!_.isString(certPath)) {
-            returncb(new Error(ERROR_MISSING_CERT_PATH));
+            return cb(new Error(ERROR_MISSING_CERT_PATH));
         }
 
-        async.mapValues([keyPath, certPath], function (value, key, cb) {
-          resolver.readFile(value, cb);
-        }, function (err, fileContents) {
+        async.map([keyPath, certPath], resolver.readFile, function (err, fileContents) {
```
czardoz: `mapValues` was correct here, you just need to call it correctly. 

1. Should pass an object to `mapValues` that looks like `{ key: keyPath, cert: certPath }`, and then access `fileContents` using `fileContents.key` and `fileContents.cert`.

lib/requester/request-wrapper.js
```diff
@@ -115,21 +115,19 @@ var _ = require('lodash'),
...
             var keyFileContents,
                 certFileContents;
 
             if (err) {
                 return cb(err);
             }
 
-            keyFileContents = fileContents[keyPath],
-            certFileContents = fileContents[certPath];
+            keyFileContents = fileContents[0],
```
czardoz: This should not be done, check above comment.

PR #197, opened 2017-03-20
Added an option to enable or disable sending of bodies with GET requests

czardoz:
Sending is disabled by default. This is a behavioral change, @shamasis

### DIFF
--- a/lib/requester/util.js
+++ b/lib/requester/util.js
@@ -1,6 +1,8 @@
 var _ = require('lodash'),
     sdk = require('postman-collection'),
-    version = require('../../package.json').version;
+    version = require('../../package.json').version,
+
+    GET = 'get';
 
 module.exports = {
 
@@ -116,15 +118,23 @@ module.exports = {
      *
      * @todo: Move this to the SDK.
      * @param request
+     * @param options
      */
-    getRequestBody: function (request) {
-        var options,
+    getRequestBody: function (request, options) {
+        var reqBody,
             mode = _.get(request, 'body.mode'),
             body = _.get(request, 'body'),
             empty = request.body && request.body.isEmpty && request.body.isEmpty(),
+            method,
             content;
 
         if (empty || !mode || !body[mode]) {
+            return;
+        }
+
+        // Do not send a body if the request is GET and sendBodyWithGetRequests is falsey.
+        method = _.isFunction(_.get(request, 'method.toLowerCase')) ? request.method.toLowerCase() : undefined;
+        if (method === GET && !options.sendBodyWithGetRequests) {
             return;
         }
 
@@ -135,10 +145,10 @@ module.exports = {
         }
 
         if (mode === 'raw') {
-            options = { body: content };
+            reqBody = { body: content };
         }
         else if (mode === 'urlencoded') {
-            options = {
+            reqBody = {
                 form: _.reduce(content, function (accumulator, param) {
                     if (param.disabled) { return accumulator; }
 
@@ -157,7 +167,7 @@ module.exports = {
             };
         }
         else if (request.body.mode === 'formdata') {
-            options = {
+            reqBody = {
                 formData: _.reduce(content, function (accumulator, param) {
                     if (param.disabled) { return accumulator; }
 
@@ -176,11 +186,11 @@ module.exports = {
             };
         }
         else if (request.body.mode === 'file') {
-            options = {
+            reqBody = {
                 body: _.get(request, 'body.file.content')
             };
         }
-        return options;
+        return reqBody;
     },
 
     /**
--- a/lib/requester/util.js
+++ b/lib/requester/util.js
@@ -5,7 +5,7 @@ var _ = require('lodash'),
     /**
      * Different modes for a request body.
      *
-     * @type {{RAW: string, URLENCODED: string, FORMDATA: string, FILE: string}}
+     * @enum {String}
      */
     REQUEST_MODES = {
         RAW: 'raw',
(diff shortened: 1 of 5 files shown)


lib/requester/util.js
```diff
@@ -116,29 +118,37 @@ module.exports = {
...
             empty = request.body && request.body.isEmpty && request.body.isEmpty(),
+            method,
             content;
 
         if (empty || !mode || !body[mode]) {
             return;
         }
 
+        // Do not send a body if the request is GET and sendBodyWithGetRequests is falsey.
+        method = _.isFunction(_.get(request, 'method.toLowerCase')) ? request.method.toLowerCase() : undefined;
```
kunagpal: https://lodash.com/docs/4.17.4#result can be used as a cleaner alternative.

czardoz: Reluctant to use it here, because it does a lot of unneeded things.

shamasis: This is okay for me. Wondering if we need to clear a body if it already exists.

czardoz: not necessary, we can let it be so that it stays saved

lib/requester/util.js
```diff
@@ -116,29 +118,37 @@ module.exports = {
...
      * @param request
+     * @param options
      */
-    getRequestBody: function (request) {
-        var options,
+    getRequestBody: function (request, options) {
+        var reqBody,
             mode = _.get(request, 'body.mode'),
             body = _.get(request, 'body'),
             empty = request.body && request.body.isEmpty && request.body.isEmpty(),
```
shamasis: the logic for getting `empty` is weird? you qlredy have `body` on top. and it returns `false` if body is undefined. Is that correct?

lib/requester/util.js
shamasis: can we pre-check `request` on top and do early bail out? When done, then this code will be

```
method = _.isString(request.method) ? request.method.toLowerCase() : undefined;
```

lib/requester/util.js
```diff
@@ -116,29 +118,37 @@ module.exports = {
      *
      * @todo: Move this to the SDK.
      * @param request
+     * @param options
      */
-    getRequestBody: function (request) {
-        var options,
+    getRequestBody: function (request, options) {
+        var reqBody,
```
shamasis: can we rename this variable to `computedBody` and move it at the end of `var` definition?

lib/requester/util.js
```diff
@@ -116,29 +118,37 @@ module.exports = {
...
+            return;
+        }
+
         content = body[mode];
 
         if (_.isFunction(content.all)) {
             content = content.all();
         }
 
         if (mode === 'raw') {
```
shamasis: make `raw` a constant

lib/requester/util.js
```diff
@@ -116,29 +118,37 @@ module.exports = {
...
-            options = { body: content };
+            reqBody = { body: content };
         }
         else if (mode === 'urlencoded') {
```
shamasis: make `urlencoded` a constant.

Better still, make an enum:
```
REQUEST_MODE = {
    RAW: 'raw',
    URLENCODED: 'urlencoded'
    FORMDATA: 'formdata'
}
```

lib/requester/util.js
```diff
@@ -1,6 +1,20 @@
 var _ = require('lodash'),
     sdk = require('postman-collection'),
-    version = require('../../package.json').version;
+    version = require('../../package.json').version,
+
+    /**
+     * Different modes for a request body.
+     *
+     * @type {{RAW: string, URLENCODED: string, FORMDATA: string, FILE: string}}
```
shamasis: What jsDoc is this?
@enum {string} should be it.

lib/requester/util.js
```diff
@@ -116,15 +130,23 @@ module.exports = {
...
      */
-    getRequestBody: function (request) {
-        var options,
-            mode = _.get(request, 'body.mode'),
+    getRequestBody: function (request, options) {
+        var mode = _.get(request, 'body.mode'),
             body = _.get(request, 'body'),
-            empty = request.body && request.body.isEmpty && request.body.isEmpty(),
-            content;
+            empty = body ? body.isEmpty() : true,
```
shamasis: Who will check function type of body? It was there previously. God knows what edge cases will break because of this. 

`_.isFunction(body && body.isEmpty) ? body.isEmpty() : true;`

lib/requester/util.js
```diff
@@ -116,15 +130,23 @@ module.exports = {
...
+            method,
+            content,
+            computedBody;
+
+        if (!request || empty || !mode || !body[mode]) {
```
shamasis: Why do || based checks? It will needlessly evaluate till end. 

Next if body is undefined checking mode in it will cause error.

lib/requester/util.js
```diff
@@ -116,15 +130,23 @@ module.exports = {
...
+            return;
+        }
 
-        if (empty || !mode || !body[mode]) {
+        // Do not send a body if the request is GET and sendBodyWithGetRequests is falsey.
+        method = _.isString(request.method) ? request.method.toLowerCase() : undefined;
```
shamasis: Why not get method in the var definition?

lib/requester/util.js
```diff
@@ -116,15 +130,23 @@ module.exports = {
...
+        if (method === GET && !options.sendBodyWithGetRequests) {
```
shamasis: Why not (after putting method in var definition, move this in early bail out

lib/requester/util.js
czardoz: Nothing will break, the extra guard was added back when we were maintaining compatibilty with an older version of the SDK which did not have this method :)

PR #205, opened 2017-03-27
Added Crypto MD5 sandbox integration tests

kunagpal:
Newman equivalent: https://github.com/postmanlabs/newman/blob/develop/test/integration/crypto-md5.postman_collection.json


czardoz:
This is already tested in the UVM spec: https://github.com/postmanlabs/postman-runtime/blob/develop/test/integration-legacy/uvm-spec.test.js#L29

Do we want to have this duplicate test?

kunagpal:
@czardoz In that case, this can be closed.

PR #216, opened 2017-03-27
Added Sandbox library integration tests

kunagpal:
Newman equivalent: https://github.com/postmanlabs/newman/blob/develop/test/integration/super-sandbox-test.postman_collection.json


czardoz:
These are copied over from Newman, and overlap heavily with the current `uvm-spec`. We should take some time, and migrate the existing UVM spec to this file.

PR #225, opened 2017-03-31
upgrade to sandbox v2, which uses pure v2 data structure

shamasis:


lib/runner/extensions/event.command.js
```diff
@@ -184,7 +184,7 @@ module.exports = {
                         // @todo find a better way to not sync entire scope, but receive changes and update the existing
                         // scope using that
                         if (_.isFunction(payload.context[variable].syncVariablesFrom)) {
-                            payload.context[variable].clear();
+                            payload.context[variable].values.clear();
```
shamasis: We did add clear to VariableScope itself. Why's this a bug?

PR #230, opened 2017-04-04
Properly update to postman-sandbox v2.x (tests and all)

czardoz:
* updated postProcessContext to process test failures correctly from the new execution object
* updated set next request logic to account for the new execution object

* fixed tests

### DIFF
--- a/lib/runner/extensions/event.command.js
+++ b/lib/runner/extensions/event.command.js
@@ -177,20 +177,20 @@ module.exports = {
                     err && (err = serialisedError(err, true));
 
                     // if it is defined that certain variables are to be synced back to result, we do the same
-                    track && result && result.globals && track.forEach(function (variable) {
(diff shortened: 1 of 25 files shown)


lib/runner/extensions/event.command.js
```diff
@@ -177,26 +177,26 @@ module.exports = {
                     err && (err = serialisedError(err, true));
 
                     // if it is defined that certain variables are to be synced back to result, we do the same
-                    track && result && result.globals && track.forEach(function (variable) {
-                        if (!(_.isObject(result.globals[variable]) && result.globals[variable])) { return; }
+                    track && result && track.forEach(function (variable) {
+                        if (!(_.isObject(result[variable]) && result[variable])) { return; }
```
shamasis: `_.isObject(result[variable]) && payload.context[variable]`

PR #235, opened 2017-04-10
ensure that we bubble up sdk instances, not serialized POJOs

czardoz:
Ensures that the Script Results are consistent with the rest of the SDK events API.

### DIFF
--- a/lib/runner/extensions/event.command.js
+++ b/lib/runner/extensions/event.command.js
@@ -1,4 +1,5 @@
 var _ = require('lodash'),
+    sdk = require('postman-collection'),
     async = require('async'),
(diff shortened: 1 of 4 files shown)


lib/runner/extensions/event.command.js
```diff
@@ -198,6 +199,13 @@ module.exports = {
                     if (!err && (abortOnFailure || stopOnFailure)) {
                         err = postProcessContext(result);
                     }
+
+                    // Ensure that we have SDK instances, not serialized plain objects.
```
kunagpal: Let's have a todo to update `postman-sandbox` to avoid serialization on every event trigger, this looks fine for now.

PR #252, opened 2017-05-08
Fixed the beforeRequest and request trigger behavior

czardoz:
Also fixed some tests.


czardoz:
@shamasis You should check this.

PR #263, opened 2017-05-15
Allowed all SSL context ciphers in request wrapper agent options

kunagpal:
Fixes https://github.com/postmanlabs/newman/issues/797

> TODO:
- [ ] Add unit test with deprecated SSL cipher


shamasis:
In cognisance to really edge-case security issue, lets do this only when `--insecure-ciphers` is set from CLI.

shamasis:
In cognisance to really edge-case security issue, lets do this only when `--insecure-ciphers` is set from CLI.

shamasis:
Pass option to runtime that takes insecureCipher http requester option.

PR #269, opened 2017-05-23
Backported the io event from feature/authorizer

czardoz:
- [x] Backported the event into the requester
- [x] Added tests
- [x] Updated the readme

### DIFF
--- a/lib/requester/requester.js
+++ b/lib/requester/requester.js
@@ -3,7 +3,23 @@ var requests = require('./request-wrapper'),
     _ = require('lodash'),
     sdk = require('postman-collection'),
 
-    HTTP = 'http',
+    /**
+     * This is a default trace object returned in all requester IO events.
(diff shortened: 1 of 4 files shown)


lib/requester/requester.js
```diff
@@ -94,7 +97,12 @@ Requester.prototype.request = function (item, cb, scope) {
         startTime = Date.now();
 
     return requests(request, requestOptions, function (err, res, resBody) {
-        if (err) { return cb.call(scope || this, err, undefined, request); }
+        if (err) {
+            // bubble up http errors
+            // @todo - Should we send an empty sdk Response here?
+            triggers.io(err, cursor, HTTP, undefined, request);
```
shamasis: @czardoz we cannot always send this as type. we need to define accurate type info that is to be sent during requesting.

PR #278, opened 2017-06-12
Added a bash script to perform memory profiling

czardoz:


npm/memory-check.sh
```diff
@@ -0,0 +1,197 @@
...
+	local TEST_SCRIPT="memtest.js";
+
+	# Enter the version directory
+	cd ${VERSION_DIR};
+
+	git checkout ${VERSION} &> /dev/null;
+
+	# Create the test script
+	cat > ${VERSION_DIR}/${TEST_SCRIPT} <<-EOF
+	var sdk = require('postman-collection'),
```
kunagpal: The code here could be take from a test template.

kunagpal: Also, an exception for this script will have to be added to the package script system tests.

czardoz: Better to leave the script self contained. Any errors will cause the memory tests to stop anyway.

czardoz: But we can do this if you are hung up on it :+1:

czardoz: Oh no, we can't.. need to be able to profile older versions too.

PR #288, opened 2017-06-20
Integrate `pm.sendRequest()` with Runtime

czardoz:
Now we can send requests from the Sandbox!


lib/authorizer/index.js
```diff
@@ -144,7 +144,7 @@ _.assign(Authorizer.prototype, /** @lends Authorizer.prototype */ {
 
         Requester.create({
             type: 'http',  // @todo - how can the auth declare what type it needs?
-            source: _.get(handler, AUTH_TYPE_PROP, POST + '.auth')
+            source: _.get(handler, AUTH_TYPE_PROP, POST) + '.auth'
```
kunagpal: Can be referenced via a `DOT_AUTH` constant.

test/integration/request-flow/requests-from-sandbox.test.js
```diff
@@ -0,0 +1,240 @@
...
+            this.run({
+                collection: {
+                    item: {
+                        // ensure that we run something for test and pre-req scripts
+                        event: [{
+                            listen: 'prerequest',
+                            script: {
+                                exec: `
+                                var sdk = require('postman-collection'),
+                                    myreq = new sdk.Request('${sandboxRequestUrl}');
```
kunagpal: Single quotes used instead of backticks.

test/integration/request-flow/requests-from-sandbox.test.js
```diff
@@ -0,0 +1,240 @@
...
+        before(function(done) {
+            this.run({
+                collection: {
+                    item: {
+                        // ensure that we run something for test and pre-req scripts
+                        event: [{
+                            listen: 'prerequest',
+                            script: {
+                                exec: `
+                                var sdk = require('postman-collection'),
```
kunagpal: Incorrect indentation

test/integration/request-flow/requests-from-sandbox.test.js
```diff
@@ -0,0 +1,240 @@
...
```
kunagpal: Incorrect indentation

lib/authorizer/index.js
```diff
@@ -144,7 +145,7 @@ _.assign(Authorizer.prototype, /** @lends Authorizer.prototype */ {
 
         Requester.create({
             type: 'http',  // @todo - how can the auth declare what type it needs?
-            source: _.get(handler, AUTH_TYPE_PROP, POST + '.auth')
+            source: _.get(handler, AUTH_TYPE_PROP, POST) + DOT_AUTH
```
shamasis: These string concats can be done on top?

lib/runner/extensions/event.command.js
```diff
@@ -75,6 +77,27 @@ module.exports = {
                 run.triggers.assertion.apply(run.triggers, arguments);
             });
 
+            context.on('execution.request', function (cursor, id, requestId, request) {
+                Requester.create({
+                    type: 'http',
+                    source: 'script',  // @todo - get this from the sandbox
```
shamasis: You need to know what script. Right?

lib/runner/extensions/event.command.js
```diff
@@ -75,6 +77,27 @@ module.exports = {
...
+                    cursor: cursor
+                }, run.options.requester, function (err, requester) {
+                    if (err) {
+                        // @todo is this correct? :-o
+                        return run.triggers.exception.apply(run.triggers, arguments);
```
shamasis: No. We should send error back inside.

lib/runner/extensions/event.command.js
```diff
@@ -75,6 +77,27 @@ module.exports = {
...
+                    }
+
+                    var sendId = id + '.' + requestId;
+
+                    requester.on(sendId, run.triggers.io.bind(run.triggers));
```
shamasis: This is a source of leak. Can we find a way to safely remove this?

lib/runner/extensions/event.command.js
```diff
@@ -75,6 +77,27 @@ module.exports = {
...
+
+                    return requester.request(sendId, new sdk.Request(request), function (err, res) {
+                        context.dispatch('execution.response.' + id, requestId, err, res && (new sdk.Response(res)));
```
shamasis: For posterity, we should use isRespose here.

lib/runner/extensions/event.command.js
czardoz: Yep. That's not available from the sandbox, we need that information.

lib/authorizer/index.js
czardoz: Uh, we might get different strings because of `_.get`

lib/runner/extensions/event.command.js
czardoz: @shamasis If there's an error and no response is available, do you want to send an empty response object inside the sandbox?

czardoz: Actually, that type checking is not required here, removing the check altogether.

czardoz:
@shamasis Ensured that `.dispose()` is called whereever needed. This PR is ready for review + merge.

PR #291, opened 2017-06-23
temporary bugfix for the auth variable resolution

czardoz:


lib/runner/request-helpers-presend.js
```diff
@@ -18,6 +18,11 @@ module.exports = [
         // Re-parse the URL, because variables have been resolved now, and things might be moved around
         item.request.url = new (sdk.Url)(item.request.url.toString());
 
+        // @todo - temp hack, this should go once we implement auth vars as a list, and start tracking which ones were
+        // added when.
+        // Basically, auth variables need to undergo the variable resolution step too.
+        _.has(item.request, 'auth') && (context.auth = item.getAuth());
```
kunagpal: Shouldn't this check be stricter?

czardoz: What should it be?

kunagpal: Ignore, as this is temporary anyway.

PR #308, opened 2017-07-04
Set awsv4 as type for exported auth handlers

kamalaknn:


kunagpal:
@kamalaknn Why was this necessary? 🤔

kamalaknn:
@kunagpal Because SDK exports AWS auth as `awsv4` [here](https://github.com/postmanlabs/postman-collection/blob/develop/lib/collection/request-auth.js#L22) and that would mean [here](https://github.com/postmanlabs/postman-runtime/blob/9baf69934d095476b16e5970b9c9cde1ebc1b793/lib/authorizer/index.js#L121) there would be a handler but that handler does not have a matching authorizer in runtime and `_sign` method would be unavailable. So it throws `handler._sign is not a function` with AWS auth.

PR #323, opened 2017-07-18
Adds concept of RequesterPool to persist options

shamasis:
The premise of this PR comes from the bug where cookie jar was not persistent across requests.

Creating new instances of Requester whenever needed caused to not be able to store common stuff across requesters. This is still a temporary measure. 

For example, a reference to cookie jar could not be re-used.

Long Term: Down the line, CookieJar store would be forwarded to the request library after embedding the requester fully into `postman-request` as such, like sandbox, moving requester completely out into a separate external module.

### DIFF
--- a/test/integration/sanity/cookie-handling.test.js
+++ b/test/integration/sanity/cookie-handling.test.js
@@ -14,6 +14,15 @@ describe('cookies', function () {
                         }
                     }],
                     request: 'http://postman-echo.com/cookies/set?foo=bar'
+                }, {
+                    event: [{
+                        listen: 'test',
+                        script: {
+                            type: 'text/javascript',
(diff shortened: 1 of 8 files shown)


test/integration/sanity/cookie-handling.test.js
```diff
@@ -24,10 +33,19 @@ describe('cookies', function () {
 
     it('must have run the test script successfully', function () {
         expect(testrun).be.ok();
-        expect(testrun.test.calledOnce).be.ok();
```
czardoz: Why should this be called twice?

shamasis: Because it has two requests?

czardoz: Oh, the test fixture's updated too. Missed that

PR #324, opened 2017-07-20
Set tunnel to true only for https URLs

harryi3t:
Ideally, tunnel should also work with http URLs ~and~
~HTTPS should also work without tunnel~

But since this is not working and the reason is not known, we will be using tunnel for only https URLs for now

Also, the tunnel option from App will be removed until this is fixed

### DIFF
--- a/lib/requester/request-wrapper.js
+++ b/lib/requester/request-wrapper.js
@@ -81,7 +81,10 @@ var _ = require('lodash'),
 
         if ((proxyConfig = _.get(request, 'proxy'))) {
             options.proxy = proxyConfig.getProxyUrl(request.url);
-            options.tunnel = proxyConfig.tunnel;
+            // TODO: figure out why tunnel is not working with http request
+            // and without tunnel why https is not working
+            // When this is fixed, App needs to show option to configure tunnel
+            options.tunnel = _.startsWith(request.url, 'https://');
         }
         cb(null, request, options);
     },


czardoz:
HTTPs cannot work without tunnelling because of SSL certificate validations

harryi3t:
@czardoz updated the comment
But now also there is the issue that http requests are not working if using tunnel
Is this also expected?

czardoz:
Yes, should not use the CONNECT method for HTTP proxying

harryi3t:
@czardoz Umm.. it is possible to use CONNECT for http proxy but it is *not necessary*.
But here it breaks if tunnel is true for http request

Let's say that **we** want to implement it this way (tunnel only for https request only) and document this limitation,
then shouldn't we remove the `tunnel` property from the `ProxyCofig` object in SDK since it of no use. And then here in runtime we can set the value depending on whether the req is https or not.

czardoz:
On Jul 20, 2017 5:15 PM, "Harendra Singh" <notifications@github.com> wrote:

@czardoz <https://github.com/czardoz> Umm.. it is possible to use CONNECT
for http proxy but it is *not necessary*.
But here it breaks if tunnel is true for http request


This what do you mean "breaks"? Does the connection get dropped?

Let's say that *we* want to implement it this way (tunnel only for https
request only) and document this limitation,
then shouldn't we remove the tunnel property from the ProxyCofig object in
SDK since it of no use. And then here in runtime we can set the value
depending on whether the req is https or not.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/postmanlabs/postman-runtime/pull/324#issuecomment-316678936>,
or mute the thread

harryi3t:
It throws error
### TUNNEL: tunneling socket could not be established, statusCode=403

czardoz:
That's what I was saying, most modern proxies will not allow CONNECT method for HTTP connections. This is so that they can cache content, static assets etc.

harryi3t:
Closing this PR since it's not a bug afterall 
@shamasis But I still think that the default behavior should be to take the tunnel option depending on the request url *unless the user wants to override it* 
Otherwise the App user can't use a custom proxy with http and https both enabled

PR #336, opened 2017-08-01
fix the infinite replay issue in auth

harryi3t:
Added `maxReplayCount` to auth to avoid infinite loop issue

### DIFF
--- a/lib/runner/extensions/request.command.js
+++ b/lib/runner/extensions/request.command.js
@@ -35,6 +35,9 @@ var _ = require('lodash'),
         // get a reference to the Auth instance from the original item, so any changes don't have to be synced back
         context.auth = payload.item.getAuth();
 
+        // keep track of no of replays of the request
+        context.replayCount = context.replayCount || 0;
+
         // if an item is not provided in the context, clone the one from the payload,
         // so that we can make any changes we need there, without mutating the collection.
         context.item = context.item || new sdk.Item(payload.item.toJSON());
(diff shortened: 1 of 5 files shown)


lib/runner/extensions/request.command.js
```diff
@@ -96,9 +100,11 @@ module.exports = {
              * @property {VariableScope} globals
              * @property {Object} data
              * @property {Boolean} replay - Indicates whether this request a replayed request
+             * @property {Number} replayCount - Number of times request was replayed
              */
             context = createContext(payload);
             replayed = context.replay;
```
czardoz: Since you're storing the count now, can you get rid of this flag? (`replayed` is true if count is greater than zero)

lib/runner/extensions/request.command.js
```diff
@@ -35,6 +35,9 @@ var _ = require('lodash'),
         // get a reference to the Auth instance from the original item, so any changes don't have to be synced back
         context.auth = payload.item.getAuth();
 
+        // keep track of no of replays of the request
+        context.replayCount = context.replayCount || 0;
```
czardoz: We should have a better name for this property IMO. It's not actually the "replay" count, it's the number of times the request went through the entire pipeline. @shamasis ?

harryi3t: how about retryCount?

harryi3t: @czardoz umm... isn't `replayCount` is always one less than "the number of times the request went through the entire pipeline"?

czardoz: Yes, if you call it that. But it introduces the concept of "replay" which might not be necessary. Personally, I think it might be good to call it `sendCount` or something, and initialize it to `1` to begin with.

harryi3t:
Not needed any more

PR #341, opened 2017-08-07
Add `_request_id` to trace to uniquely identify requests

kamalaknn:
This PR adds `_request_id` to trace for each request that is triggered from `script` and `collection`. 

This allows to uniquely identify individual requests even when the `cursor` is still the same.

`_request_id` is not added to requests triggered from `Authorizer` yet. @shamasis should we create new ids for intermediate requests sent by `Authorizer`, or is there any other identifier we can use?


kamalaknn:
Superceded by #345

PR #345, opened 2017-08-08
Queue `pm.sendRequest` through `request` command

kamalaknn:
Queue `pm.sendRequest` through `request` command instead of sending the request over directly. This handles variable, proxy, and certificate resolution.

### DIFF
--- a/lib/runner/extensions/event.command.js
+++ b/lib/runner/extensions/event.command.js
@@ -75,23 +75,6 @@ module.exports = {
                 run.triggers.assertion.apply(run.triggers, arguments);
             });
 
-            context.on('execution.request', function (cursor, id, requestId, request) {
-                run.requester.create({
-                    type: 'http',
-                    source: 'script', // @todo - get script type from the sandbox
-                    cursor: cursor
-                }, function (err, requester) { // eslint-disable-line handle-callback-err
-                    var sendId = id + '.' + requestId;
-
-                    requester.on(sendId, run.triggers.io.bind(run.triggers));
-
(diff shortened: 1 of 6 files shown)


lib/runner/extensions/event.command.js
```diff
@@ -194,6 +177,27 @@ module.exports = {
...
+                var run = this;
+
+                this.host.on('execution.request', function (cursor, id, requestId, request) {
+                    // clone cursor and add source
+                    var cursorWithSource = _.clone(cursor);
+
+                    cursorWithSource.source = 'script'; // @todo - get script type from the sandbox
+
+                    run.immediate('request', {
+                        item: new sdk.Item({
```
kamalaknn: @shamasis when creating a new `Item` without a parent reference to any collection, the collection variable resolution flow [here](https://github.com/postmanlabs/postman-runtime/blob/develop/lib/runner/request-helpers-presend.js#L12) breaks. 

Unsure how to handle this.

lib/runner/extensions/event.command.js
```diff
@@ -194,6 +177,27 @@ module.exports = {
                         !assertion.passed && assertionFailed.push(assertion.name);
                     });
 
+                var run = this;
+
+                this.host.on('execution.request', function (cursor, id, requestId, request) {
+                    // clone cursor and add source
+                    var cursorWithSource = _.clone(cursor);
+
+                    cursorWithSource.source = 'script'; // @todo - get script type from the sandbox
```
kamalaknn: Needs better naming. I could think of `cursor.context` meaning if the context is `collection` or a `script`, but we have used `context` in other places to make it misleading.

PR #351, opened 2017-08-14
Prevent file uploads in `pm.sendRequest`

kamalaknn:


lib/runner/extensions/event.command.js
```diff
@@ -28,6 +33,60 @@ var _ = require('lodash'),
...
+        // do nothing if request body is empty
+        if (!request.body) {
+            return request;
+        }
+
+        // handle request body in `file` mode
+        if (request.body.mode === REQUEST_BODY_MODE_FILE) {
+            // clone the request and remove file in body
+            sanitizedRequest = _.clone(request);
+            delete sanitizedRequest.body.file;
```
kamalaknn: Unsure if `delete sanitizedRequest.body.file` is better or `sanitizedRequest.body.file = {}` is better?

shamasis: Please set the value to `null`

kamalaknn: Got it 👍

lib/runner/extensions/event.command.js
```diff
@@ -28,6 +33,60 @@ var _ = require('lodash'),
...
+        // handle request body in `formdata` mode
+        if (request.body.mode === REQUEST_BODY_MODE_FORMDATA) {
+            // extract any file type params
+            sanitizedFormdata = _.reject(request.body.formdata, {type: FILE});
+
+            // if a file param was detected
+            if (sanitizedFormdata.length !== _.get(request, 'body.formdata.length')) {
+                // clone the request and set form data without file params
+                sanitizedRequest = _.clone(request);
+                sanitizedRequest.formdata = sanitizedFormdata;
```
kunagpal: `sanitisedRequest[REQUEST_BODY_MODE_FORMDATA] = sanitizedFormdata;`

lib/runner/extensions/event.command.js
```diff
@@ -28,6 +33,60 @@ var _ = require('lodash'),
...
+
+            // send a warning that file was removed
+            warn();
+
+            return sanitizedRequest;
+        }
+
+        // handle request body in `formdata` mode
+        if (request.body.mode === REQUEST_BODY_MODE_FORMDATA) {
+            // extract any file type params
```
kunagpal: I feel this segment could be made cleaner by:
1. Cloning the request.
2. Traversing the members of the cloned formdata request body.
3. Setting any encountered `src` values to null.

This ensures that at most one copy+traversal of the request/request-body is made. Currently, the request body has to be traversed once during `_.reject`, and then a second time via clone.

kamalaknn: @kunagpal the reason I do `_.reject` first is, not all requests have file types in form data.

We should do a `clone` only when we detect a file param. Otherwise, we would be doing the `clone` even for requests without `files` which is costlier than `_.reject` of `formdata` which would be much smaller.

lib/runner/extensions/event.command.js
kamalaknn: Will change 👍 

Is there any particular benefit of this or code quality in general?

kunagpal: Should there ever be a change in the name given to formdata bodies, we'll have to change it in just one place.

lib/runner/extensions/event.command.js
kunagpal: Fair enough.

lib/runner/extensions/event.command.js
```diff
@@ -28,6 +33,60 @@ var _ = require('lodash'),
...
+    /**
+     * Removes files in Request body if any.
+     *
+     * @param {Request~definition} request Request JSON representation to be sanitized
+     * @param {Function} warn function to invoke when a file is found
+     *
+     * @returns {Request~definition}
+     */
+    sanitizeFiles = function (request, warn) {
+        var sanitizedRequest,
```
kamalaknn: Will do 👍

lib/runner/extensions/event.command.js
kamalaknn: Ah okay, makes sense.

lib/runner/extensions/event.command.js
```diff
@@ -28,6 +33,56 @@ var _ = require('lodash'),
             error.name = ASSERTION_FAILURE;
         }
         return error ? serialisedError(error, true) : undefined;
+    },
+
+    /**
+     * Removes files in Request body if any.
+     *
```
shamasis: `@private` this please

kunagpal:
@kamalaknn Any specific reason for returning the sanitized request body via a callback?

lib/runner/extensions/event.command.js
```diff
@@ -28,6 +33,56 @@ var _ = require('lodash'),
...
+     * @param {Request~definition} request Request JSON representation to be sanitized
+     * @param {Function} callback function invoked with error and sanitizedRequest.
+     *                            sanitizedRequest is `null` if no files are detected.
+     */
+    sanitizeFiles = function (request, callback) {
```
shamasis: I think callback should always get the request from parameter so that it doesn't have to refer to the outer closure. ... as such the signature of callback could be `error:Error, request:Request, affectedEntries:Array.<src:String>` - in this way, we would know that nothing was affected if `_.isEmpty(affectedEntries) = true`.

lib/runner/extensions/event.command.js
```diff
@@ -183,33 +238,52 @@ module.exports = {
...
+                    // remove files in request body if any
+                    sanitizeFiles(request, function (err, sanitizedRequest) {
+                        // @todo - there is no error flows now
+                        // this is here to allow extending sanitizeFiles
+                        // to include error flows future
+                        if (err) {
+                            return;
+                        }
+
+                        var tempRequest = request,
```
shamasis: we would not need `tempRequest` variable if sanitiseFIles always return a request or null if request itself is missing / cannot be constructed.

lib/runner/extensions/event.command.js
kamalaknn: @shamasis I have a question.

If callback always gets the request, raw and mutated, how is that different from the synchronous version with the return?

Also, the signature of `affectedEntries` is not always an `Array`, it could the entire body for `file` model. If I had to send the raw request and affected entries, and let the callback to the reconstruction of new request then `sanitizeFiles` would not do much and it leaks concerns, right?

shamasis:
```
var outerFunction = function (request, callback) {
   innerFunction(request, function (request) {
      if (!request) { return callback(new Error()); }
      callback(null, request);
   });
};

var innerFunction = function (callback) {
};

// we can then write
outerFunction(myReqFromSandbox, function (err, ...) {});
```

lib/runner/extensions/event.command.js
```diff
@@ -183,33 +257,48 @@ module.exports = {
                     });
 
                 this.host.on(EXECUTION_REQUEST_EVENT_BASE + executionId, function (cursor, id, requestId, request) {
-                    // clone cursor and add source
-                    var cursorWithSource = _.clone(cursor),
+                    // remove files in request body if any
+                    sanitizeFiles(request, function (err, sanitizedRequest, affectedEntries) {
```
shamasis: call this parameter `request` itself so that it overrides the outer closure variable in request and prevent bigs.

kamalaknn: I renamed it so linter does not complain about duplicate variable names. 

Will change.

PR #361, opened 2017-08-23
Make sure empty Urls do not break runs

kamalaknn:
`util.getRequestOptions` internally called `url.getHost` which errored out for empty host. `url.getHost` should not have errored in the first place, raised a [PR](https://github.com/postmanlabs/postman-collection/pull/443) for that as well.

Plus it doesn't make sense to proceed with `requester.request` when the `url` is empty. Hence added a bail out for this case.

### DIFF
--- a/lib/requester/requester.js
+++ b/lib/requester/requester.js
@@ -101,7 +101,7 @@ _.extend(Requester.prototype, /** @lends Requester.prototype */ {
         // at this point the request could have come from collection, auth or sandbox
         // we can't trust the integrity of this request
         // bail out if request url is empty
-        if (!(request && request.url && request.url.toString())) {
+        if (!(request && request.url && request.url.toString && request.url.toString())) {
             error = new Error('runtime:extenstions~request: request url is empty');
             this.emit(id, error, this.trace.cursor, this.trace, undefined, request);
             return callback(error, undefined, request);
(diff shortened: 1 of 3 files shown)


lib/requester/requester.js
```diff
@@ -92,10 +92,24 @@ _.extend(Requester.prototype, /** @lends Requester.prototype */ {
...
+        var error,
+            self = this,
+            cookieJar,
+            requestOptions,
+            startTime;
+
+        // at this point the request could have come from collection, auth or sandbox
+        // we can't trust the integrity of this request
+        // bail out if request url is empty
+        if (!(request && request.url && request.url.toString())) {
```
kunagpal:
```js
    if (!(request && request.url && request.url.toString && request.url.toString())) {
```

kamalaknn: @kunagpal `toString` is on the `Object.prototype` which means it should be available on all types except `undefined` and `null` which `!request.url` handles. The only case where this might be otherwise is when someone does `delete url.toString` or `url.toString = null`. In that case, we may need to check `_.isFunction`.

Isn't that too much work on the critical path? I may be wrong as well.

kunagpal: The suggested change was indeed for `null` and `undefined`.

kamalaknn: Updated 👍

PR #365, opened 2017-08-30
Updated auth control flows

kamalaknn:
1. Replays and intermediate requests now first class.
2. More granularity in auth hook responsibilities
3. Remove `Authorizer` as an orchestrator
4. Introduce `ReplayController` to handle replays/intermediate requests


lib/runner/replay-controller.js
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Sends a request in the item. This takes care of limiting the total number of replays for a request.
+     * 
+     * @param {Object} context
+     * @param {Request} item
+     * @param {Boolean} waitForResponse if true, callback waits for request to complete
+     * @param {Function} success this callback is invoked when replay controller sent the request
+     * @param {Function} failure this callback is invoked when replay controller decided not to send the request
+     */
+    requestReplay: function (context, item, waitForResponse, success, failure) {
```
kamalaknn: @shamasis I'm not 100% confident with the interface for `requestReplay`.

lib/runner/request-helpers-presend.js
```diff
@@ -18,32 +20,92 @@ module.exports = [
...
+                    });
+                }, function () {
+                    // replay count exceeded, proceed with request
+                    done();
+                });
+            });
+        }
+
+        // start the by calling the pre hook of the auth
+        authPreHook();
```
kamalaknn: @shamasis this has become a bit complex now.

lib/runner/extensions/request.command.js
```diff
@@ -29,43 +30,46 @@ var _ = require('lodash'),
...
+        parent && (item.setParent(parent));
+
+        // save the cloned item in context
+        context.item = item;
 
-        // if an item is not provided in the context, clone the one from the payload,
-        // so that we can make any changes we need there, without mutating the collection.
-        context.item = context.item || new sdk.Item(payload.item.toJSON());
+        // save original item for reference
+        context.originalItem = item;
```
harryi3t: don't you mean payload.item?

lib/runner/extensions/request.command.js
```diff
@@ -29,43 +30,46 @@ var _ = require('lodash'),
...
+
+        // get a reference to the Auth instance from the original item, so any changes don't have to be synced back
+        !context.auth && (context.auth = context.item.getAuth());
```
harryi3t: the comment says `from the original item` but this is using the clone item

kamalaknn: It shouldn't have been `original item` in the comment. updating it.

lib/runner/extensions/request.command.js
kamalaknn: It should have been `payload.item` but, it won't have `parent` which makes it difficult for replays.

lib/runner/replay-controller.js
```diff
@@ -0,0 +1,105 @@
+var _ = require('lodash'),
+
+    // total number of replays allowed
+    MAX_REPLAY_COUNT = 3,
+
+    ReplayController;
+
+/**
+ * Handles replay logic with. Makes sure request replays do not go into an infinite loop.
```
harryi3t: with?

lib/runner/extensions/request.command.js
```diff
@@ -29,43 +30,46 @@ var _ = require('lodash'),
...
+        parent && (item.setParent(parent));
+
+        // save the cloned item in context
+        context.item = item;
 
+        // save original item for reference
+        context.originalItem = payload.item;
+
+        // get a reference to the Auth instance from the item, so any changes don't have to be synced back
+        !context.auth && (context.auth = context.item.getAuth());
```
harryi3t: @shamasis for the case of folder level auth,
- original item's parent will have the auth
- after we clone the auth and set the parent, it's done using reference 
- getAuth also returns a reference
- so now if we change the context.auth, wouldn't it change the original item's parent's auth also?

kamalaknn: @harryi3t updated code to resolve in `presend` helpers. When variables are resolved, auth is cloned and updating the auth from then on will not update the original item.

Of course, there is a possibility of mutating auth at the time between creating context and resolving variables in `presend`

lib/runner/extensions/request.command.js
```diff
@@ -29,43 +30,49 @@ var _ = require('lodash'),
...
+        context.item = item;
+
+        // save original item for reference
+        context.originalItem = payload.item;
+
+        // get a reference to the Auth instance from the item, so any changes don't have to be synced back
+        !context.auth && (context.auth = context.item.getAuth());
+
+        // set parent of auth to request so that SDK can use it for variable resolution
+        auth.setParent(context.originalItem.request);
```
kamalaknn: @shamasis should this be done in SDK as part of `Request` constructor?

shamasis:
Merging this for a sane review process.

PR #367, opened 2017-08-31
Allow username, password to be empty string in basic auth

harryi3t:
Support Issue: https://github.com/postmanlabs/postman-app-support/issues/3458


lib/authorizer/basic.js
```diff
@@ -43,7 +44,7 @@ module.exports = {
      * @todo - this should be made into ".sign" after the signature calculation is moved out of the SDK.
      */
     sign: function (params, request) {
-        if (!params || !params.username) { return request; } // Nothing to do if no parameters are present.
+        if (!params || !_.has(params, 'username') || !_.has(params, 'password')) { return request; }
```
harryi3t: instead of checking for truthy value (since the empty string is not truthy, it was bailing out), we now only check for the existence of the property
Should we also check the type to be a string?

kunagpal: This condition will still cause a bailout if the username is not present. Since username is not meant to be required anymore, you'd want to check for the password exclusively, like so:
```js
if (!(params && _.has(params, 'password'))) { return request; }
```

test/unit/authorizer.test.js
```diff
@@ -39,41 +39,46 @@ describe('Authorizers', function () {
...
+            rawBasicReq.auth.basic = {password: 'foo'}; // no username present
+            request = new Request(rawBasicReq);
+            auth = request.auth;
+            handler = Authorizer.Handlers[auth.type];
+            authorizedReq = handler.sign(auth.parameters().toObject(), request);
+
+            expect(authorizedReq.headers.all()).to.eql([]);
+        });
+
+        it('should work with empty username and password', function () {
```
kunagpal: Also add a test for a missing username.

lib/authorizer/basic.js
kunagpal: Empty strings are still valid, property existence checks will suffice here.

harryi3t: yeah, this PR makes empty string valid!
I was asking whether we should add a check that the username and password parameter MUST be a string (not an object or Array)

test/unit/authorizer.test.js
harryi3t: line 57

lib/authorizer/basic.js
harryi3t: No no no, this does not allow the username to be absent
Only to be empty string (same for password, it can be empty string)
I will change the title to be more precise

kunagpal: Yes, do add the string like check.

lib/authorizer/basic.js
```diff
@@ -42,7 +43,9 @@ module.exports = {
      * @param {Request} request
      */
     sign: function (params, request) {
-        if (!params || !params.username) { return request; } // Nothing to do if no parameters are present.
+        if (!(params && _.isString(params.username) && _.isString(params.password))) {
```
shamasis: down below in line 53, `btoa(params.username + ':' + params.password)` should change to:

```
btoa((params.username || '') + ':' + (params.password || ''))
```

Otherwise `undefined` as a string will become the username or password!!

harryi3t: ^ @kunagpal

test/unit/authorizer.test.js
```diff
@@ -28,52 +29,56 @@ describe('Authorizers', function () {
...
                 authHeader;
             expect(headers.length).to.eql(1);
 
             authHeader = headers[0];
-            expect(authHeader.toString()).to.eql('Authorization: Basic YWJoaWppdDprYW5l');
+            expect(authHeader.toString()).to.eql(expectedAuthHeader);
             expect(authHeader.system).to.be(true);
         });
 
-        it('should bail out if username is missing', function () {
```
harryi3t: now if something is missing we make default it to empty string

kunagpal: You could retain the test and assert that the fallback is indeed an empty string. (If not done already)

harryi3t: that is handled in the test case `should use default values for the missing parameters`

harryi3t: line 80

PR #373, opened 2017-09-11
Refactor replay mechanism and introduce `response` callback

kamalaknn:
Pending

 - [x] Docs, inline and Readme
 - [x] Tests for `response` callback


lib/runner/extensions/request-wrapper.command.js
```diff
@@ -0,0 +1,32 @@
+var _ = require('lodash');
+
+module.exports = {
+    init: function (done) {
+        done();
+    },
+
+    triggers: ['response'],
+
+    process: {
```
shamasis: rename this to `request` and current `request` to `http-request`

lib/runner/extensions/request-wrapper.command.js
```diff
@@ -0,0 +1,32 @@
...
+        requestWrapper: function (payload, next) {
+            var abortOnError = _.has(payload, 'abortOnError') ? payload.abortOnError : this.options.abortOnError,
+                complete = function (nextPayload, err) {
+                    this.triggers.response(
```
shamasis: resolve variables here

lib/runner/extensions/request-wrapper.command.js
```diff
@@ -0,0 +1,32 @@
...
+                        nextPayload.coords,
+                        nextPayload.response,
+                        nextPayload.request,
+                        payload.item,
+                        nextPayload.cookies
+                    );
+                    next(err && abortOnError ? err : null, nextPayload, err);
+                }.bind(this);
+
+            this.immediate('request', payload)
```
shamasis: why not just queue?

lib/runner/extensions/request.command.js
```diff
@@ -190,9 +205,10 @@ module.exports = {
                         context.response = res;
 
                         // run the post request helpers, which need to use the response, assigned above
-                        async.applyEachSeries(posthelpers, context, self, function (/* err */) {
+                        async.applyEachSeries(posthelpers, context, self, function (err, options) {
+                            if (err) { /** */ }
                             // @todo: need to handle err
-                            complete();
+                            complete(options[0]);
```
shamasis: need to explain what this is.

lib/runner/extensions/request.command.js
```diff
@@ -190,9 +205,10 @@ module.exports = {
                         context.response = res;
 
                         // run the post request helpers, which need to use the response, assigned above
-                        async.applyEachSeries(posthelpers, context, self, function (/* err */) {
+                        async.applyEachSeries(posthelpers, context, self, function (err, options) {
+                            if (err) { /** */ }
```
shamasis: handle error here? question is, what nature of error will bubble up here and then are they supposed to stop run?

kamalaknn: Handled error and enabled test cases for asserting the error.

lib/runner/extensions/request-wrapper.command.js
kamalaknn: Added docs to explain why.

lib/runner/extensions/request-wrapper.command.js
kamalaknn: Not trivial to include in this PR. Will do it separately.

lib/runner/replay-controller.js
```diff
@@ -55,45 +54,33 @@ _.assign(ReplayController.prototype, /** @lends ReplayController.prototype */{
...
-        // or schedule a request and continue with current queue
-        else {
-            this.run.interrupt('request', payload);
-            success();
-        }
+        this.run.immediate('httpRequest', payload)
+            .done(function (response) {
+                success(null, response);
+            })
+            .catch(function (err) {
```
kunagpal: Shouldn't this be a direct reference to `failure`?

lib/runner/request-helpers-postsend.js
```diff
@@ -23,23 +24,17 @@ module.exports = [
...
-            // Auth wants to replay the request
-            var replayController = new ReplayController(context.replayState, run);
-
-            // finish up current request and schedule a replay
-            replayController.requestReplay(context, context.item, false, function (err) {
-                done(err);
-            }, function (err) {
-                done(err);
-            });
+            if (success) { return done(); }
```
kunagpal: Shouldn't this be an explicit `replay: false`?

lib/runner/replay-controller.js
kamalaknn: If you are talking about `success` vs `failure`, here `success` is when `ReplayController` decides to send a request. `failure` is when it decided not to replay the request due to max retries exceeding and not the failure in the actual request.

lib/runner/request-helpers-postsend.js
kamalaknn: IMHO, the replays should be an opt-in flow. I am not sure about forcing every other flow to turn off the. The way I see it, I want more options to be introduced. In that case wouldn't it be too much of an overhead?

kamalaknn:
@kunagpal Overall this is not a breaking change. We are introducing a new callback without touching existing ones. The tests concur.

The only change to existing flows (as can be seen from changes in tests) is, errors in `post helpers` are now bubbled up. This I think needed to be fixed.

lib/runner/replay-controller.js
kunagpal: Either way, `success(err);` looks weird.

lib/runner/extensions/event.command.js
```diff
@@ -258,7 +258,7 @@ module.exports = {
 
                         cursorWithSource._source = 'script'; // @todo - get script type from the sandbox
 
-                        this.immediate('request', {
+                        this.immediate('httpRequest', {
```
shamasis: case insensitive - all lowercase

PR #375, opened 2017-09-12
Allow updation of user properties when the value is empty string

harryi3t:
This will also convert the property to system property (will add system: true)

This is needed for 2 reasons
 - App sets the parameters as empty string which are not set by the user
 - Exported collections have parameters as empty string

### DIFF
--- a/lib/authorizer/auth-interface.js
+++ b/lib/authorizer/auth-interface.js
@@ -1,4 +1,5 @@
 var _ = require('lodash'),
+    EMPTY = '',
     createAuthInterface;
 
 /**
@@ -70,7 +71,7 @@ createAuthInterface = function (auth) {
                 }
 
                 // Update if the param is a system property or an empty user property (null, undefined or empty string)
-                if (param.system || param.value === '' || _.isNil(param.value) || _.isNaN(param.value)) {
+                if (param.system || param.value === EMPTY || _.isNil(param.value) || _.isNaN(param.value)) {
                     return param.update({key: key, value: value, system: true});
                 }
             });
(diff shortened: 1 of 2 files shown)


lib/authorizer/auth-interface.js
```diff
@@ -66,7 +66,13 @@ createAuthInterface = function (auth) {
                 var param;
                 // if auth has the param already then update it only when it's a sytem property
                 if ((param = parameters.one(key))) {
-                    param.system && param.set(value);
+                    if (param.system) {
+                        param.set(value);
```
kunagpal:
```js
param.set(value);
!param.value && (param.system = true);
```

lib/authorizer/auth-interface.js
```diff
@@ -66,7 +66,13 @@ createAuthInterface = function (auth) {
...
+                    }
+                    else if (param.value === '') {
```
kunagpal: `!param.value`

kunagpal:
@harryi3t Looks good, we could consider using `upsert` here.

test/unit/auth-interface.test.js
```diff
@@ -90,6 +91,23 @@ describe('AuthInterface', function () {
         expect(authInterface.get('nonce')).to.be(NONCE);
     });
 
+    it('should update user parameters with falsy value', function () {
+        var fakeAuthObj,
+            fakeAuth,
+            authInterface,
+            valuesToTestWith = ['', null, undefined, false, 0];
```
kamalaknn: @harryi3t are you sure you want to override `false` or 0?

Or should we `value === '' || _.isNil(value)` ?

kamalaknn:
@kunagpal we could do `upsert` but splitting it up gives more space to explain the conditions and using `upsert` here does not have a performance advantage as well.

PR #381, opened 2017-09-14
enable 'no-undef' rule in eslint

harryi3t:
Also added the globals to avoid false-positives in their folders

| folder | global rules added |
|-------|--------------------|
| npm  | cd, colors, config, exec, exit, mkdir, next, popd, pushd, rm, test |
| test | expect  |

This reverts the commit 70d1164590c6d72098edc38fbea2508bbce102d4.


.eslintrc
```diff
@@ -9,6 +9,21 @@
         "node": true,
         "es6": true
     },
+    "globals": {
```
shamasis: put it in `test/.eslintrc`

PR #382, opened 2017-09-14
Added script timeout option

kunagpal:
Sort of fixes https://github.com/postmanlabs/newman/issues/1189


lib/runner/extensions/event.command.js
```diff
@@ -284,7 +284,7 @@ module.exports = {
                 this.host.execute(event, {
                     id: executionId,
                     // debug: true,
-                    timeout: payload.scriptTimeout,
+                    timeout: payload.scriptTimeout || _.get(this.options, 'timeout.script'),
```
shamasis: Need to be _.has payload.scriptTimeout ? Payload : options

shamasis: add docs to payload

shamasis: add a todo to allow individual script level timeout

shamasis: need to follow timeout inheritence like that of request

```
timeout: _.min([
            _.get(options, 'timeout.request'),
            _.get(options, 'timeout.global')
        ]), // validated later inside requester
```

and maybe part config done in init of createContext of sandbox

kunagpal:
Blocked on a patch release of https://github.com/postmanlabs/postman-sandbox/pull/193

lib/runner/extensions/event.command.js
```diff
@@ -111,6 +111,7 @@ module.exports = {
         }
 
         sandbox.createContext(_.merge({
+            timeout: _(run.options.timeout).pick(['script', 'global']).values().min()
```
shamasis: fancy! :-)

kunagpal: I was probably in the Python one liner mode here. 😄

PR #384, opened 2017-09-15
Implement Oauth-2 and add test cases

harryi3t:


lib/authorizer/oauth2.js
```diff
@@ -1,3 +1,10 @@
+var HEADER = 'header',
+    URL = 'url',
+    DEFAULT_ADD_TOKEN_TO = HEADER,
```
kunagpal: ?

lib/authorizer/oauth2.js
```diff
@@ -44,6 +51,33 @@ module.exports = {
      * @param {AuthHandlerInterface~authSignHookCallback} done
      */
     sign: function (auth, request, done) {
+        var params = auth.get([
```
kunagpal: Shift the array to a const above.

lib/authorizer/oauth2.js
```diff
@@ -22,7 +29,7 @@ module.exports = {
      * @param {AuthHandlerInterface~authPreHookCallback} done
      */
     pre: function (auth, done) {
-        done(null, true);
+        done(null, Boolean(auth.get('accessToken')));
```
kunagpal: Consider implementing `auth.has` to avoid repetitive `Boolean` casts.

test/unit/auth-handlers.test.js
```diff
@@ -418,15 +418,98 @@ describe('Auth Handler:', function () {
...
+                    accessToken: '123456789abcdefghi',
+                    addTokenTo: 'header',
+                    tokenType: 'bearer'
+                }
+            },
+            url: 'https://api.github.com/user/orgs',
+            method: 'GET'
+        };
+
+        it('should sign the request by adding the token to the header', function () {
```
kunagpal: Also, add the corresponding logic and unit test for adding params to the request body.

lib/authorizer/oauth2.js
harryi3t: a token can be added to the header or to the URL
the default value will be `header`, if not specified on [line 64](https://github.com/postmanlabs/postman-runtime/pull/384/files#diff-7385a12d89075d66ee2143cd2e242ba5R64)

kunagpal: So use `HEADER` there instead.

test/unit/auth-handlers.test.js
harryi3t: params: see [line 448](https://github.com/postmanlabs/postman-runtime/pull/384/files#diff-ae634fae9ced1b620809db816fdab229R448)
body: not supported :(

lib/authorizer/oauth2.js
harryi3t: the check is not universal and can depend on the Auth and the parameter itself

test/unit/auth-handlers.test.js
kunagpal: https://github.com/postmanlabs/postman-app-support/issues/3239

harryi3t: The said issue is about sending the `client_id` and `client_secret` in the header/body as part of 2nd request in Oauth2 to get the `access_token`
This PR is about using `access_token` for a protected resource

The rules for the token itself are stated here https://tools.ietf.org/html/rfc6750#section-2.1
Although bearer token can also be sent in the body, the RFC says that supporting it in the header is a MUST for server and have many conditions for a client to be able to add in the body.

lib/authorizer/oauth2.js
```diff
@@ -1,3 +1,14 @@
+var HEADER = 'header',
+    URL = 'url',
```
kamalaknn: Consider `param` or `urlParam`

PR #385, opened 2017-09-15
Make sure cursor is present in all the callbacks

kamalaknn:


lib/runner/extensions/http-request.command.js
```diff
@@ -144,7 +144,11 @@ module.exports = {
                     // Since we encountered an error before even attempting to send the request, we bubble it up
                     // here.
                     afterRequest(err, undefined, item.request);
-                    return safeNext(err, {request: item.request}, err);
+                    return safeNext(
+                        err,
```
kunagpal: Why is `err` being passed twice?

test/integration/request-flow/intermediate-requests.test.js
```diff
@@ -76,9 +76,11 @@ describe('intermediate requests from auth', function () {
 
         it('must have sent the original request', function () {
             var err = testrun.response.firstCall.args[0],
+                cursor = testrun.response.firstCall.args[1],
                 request = testrun.response.firstCall.args[3];
 
             expect(err).to.be(null);
+            expect(cursor).to.not.be(null);
```
kunagpal: Would it be possible to write a stricter assertion?

kamalaknn: A stricter assertion would be to check the shape of cursor like `expect(testrun.response.firstCall.args[1]).to.have.property('ref');`. Asserting on values is not feasible.

lib/runner/extensions/http-request.command.js
kamalaknn: Good catch, we had to pass `err` twice to `next`. Now that is handled by `safeNext`.

PR #386, opened 2017-09-15
Implement bearer-token auth and add test cases

harryi3t:


lib/authorizer/index.js
```diff
@@ -81,6 +81,7 @@ _.forEach({
     noauth: require('./noauth'),
     awsv4: require('./aws4'),
     basic: require('./basic'),
+    bearerToken: require('./bearer-token'),
```
kamalaknn: Should be `bearer` or `bearertoken`.

lib/authorizer/index.js
```diff
@@ -81,7 +81,7 @@ _.forEach({
     noauth: require('./noauth'),
     awsv4: require('./aws4'),
     basic: require('./basic'),
-    bearerToken: require('./bearer-token'),
+    bearer: require('./bearer-token'),
```
kunagpal: Maintain consistency between the key and require argument.

lib/authorizer/bearer-token.js
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Initializes an item (extracts parameters from intermediate requests if any, etc)
+     * before the actual authorization step
+     *
+     * @param {AuthInterface} auth
+     * @param {Response} response
+     * @param {AuthHandlerInterface~authInitHookCallback} done
+     */
+    init: function (auth, response, done) {
+        done(null);
```
kunagpal: Why the explicit `null`?

PR #388, opened 2017-09-17
Warn on errors in auth flow without breaking request flow

kamalaknn:


lib/runner/extensions/http-request.command.js
```diff
@@ -226,7 +226,6 @@ module.exports = {
 
                         // run the post request helpers, which need to use the response, assigned above
                         async.applyEachSeries(posthelpers, context, self, function (error, options) {
-                            // bucket options per helper
```
kunagpal: Why was the comment removed? :/

lib/runner/request-helpers-postsend.js
```diff
@@ -24,13 +24,28 @@ module.exports = [
             authHandler = AuthLoader.getHandler(auth.type),
             authInterface = createAuthInterface(auth);
 
+        // bail out if there is not matching auth handler for the type
```
kunagpal: *no

lib/runner/request-helpers-presend.js
```diff
@@ -48,13 +50,31 @@ module.exports = [
             authInterface,
             authSignHook = function () {
                 try {
-                    authHandler.sign(authInterface, context.item.request, function (err) { done(err); });
+                    authHandler.sign(authInterface, context.item.request, function (err) {
```
kunagpal: In the future, `authHandler` should be adjusted to never throw, so that `try`-`catch` can be avoided.

lib/runner/extensions/http-request.command.js
kamalaknn: It wasn't relevant to the code. It shouldn't have been there in the first place.

lib/runner/request-helpers-presend.js
kamalaknn: We might not be able to do it. In the future, this part is going to be out of runtime, and we can't trust it. So, we'll have to retain it forever.

kunagpal: Alright, we'll have to add a `finally` block so that optimization bailout does not occur.

PR #394, opened 2017-09-20
Support turning off the interactive mode at individual request level

harryi3t:


test/integration/auth-methods/digest.test.js
```diff
@@ -87,6 +89,69 @@ describe('digest auth', function () {
...
+                        name: 'DigestAuth',
+                        request: {
+                            url: 'https://postman-echo.com/digest-auth',
+                            auth: {
+                                type: 'digest',
+                                digest: {
+                                    algorithm: 'MD5',
+                                    username: '{{uname}}',
+                                    password: '{{pass}}',
+                                    nonInteractive: true
```
kunagpal: 1. The schema will have to be updated to reflect this as well.
2. Add this test for all other applicable auth modes as well.

harryi3t: 1. will do
2. already done (digest auth and NTLM auth only)

lib/authorizer/digest.js
```diff
@@ -1,4 +1,6 @@
 var crypto = require('crypto-js'),
+
+    NON_INTERACTIVE = 'nonInteractive',
```
kamalaknn: IMO, now that this flag is not generic but specific to individual auths, it should not be called `nonInteractive`.

harryi3t: @kamalaknn what should it be called?

shamasis: These should be like @kamalaknn says - specific to each auth
digest may have some terminology in the RFC

lib/authorizer/ntlm.js
```diff
@@ -83,6 +84,11 @@ module.exports = {
      * @param {AuthHandlerInterface~authPostHookCallback} done
      */
     post: function (auth, response, done) {
+        // bail out if interactive mode is disabled
```
shamasis: again, this has to be NTLM specific property name

lib/authorizer/digest.js
kamalaknn: The RFC also uses the retry in its wording.

```
The client is expected to retry the request, passing an Authorization header line
```

It should be `disableRetry{something}`.

harryi3t: @kamalaknn Same is true for NTLM also
https://msdn.microsoft.com/en-us/library/cc237544.aspx
image

PR #400, opened 2017-09-22
Fix digest auth (MD5-sess)

harryi3t:
Fixes Github Issue: https://github.com/postmanlabs/postman-app-support/issues/2941

### DIFF
--- a/lib/authorizer/digest.js
+++ b/lib/authorizer/digest.js
@@ -191,7 +191,7 @@ module.exports = {
         }
 
         if (qop === 'auth' || qop === 'auth-int' || algorithm === 'MD5-sess') {
-            nonceCount && nonceCountheaderParams.push('nc=' + nonceCount);
+            nonceCount && headerParams.push('nc=' + nonceCount);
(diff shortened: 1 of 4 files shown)


kamalaknn:
According to the RFC, algorithm directive is optional and a server should ignore any directive it does not understand. 

The addition of `algorigthm="MD5"` to authorization header should not break any use cases.

lib/authorizer/digest.js
```diff
@@ -184,12 +184,14 @@ module.exports = {
...
 
+        algorithm && headerParams.push('algorithm="' + algorithm + '"');
+
         if (qop === 'auth' || qop === 'auth-int') {
             headerParams.push('qop=' + qop);
         }
 
         if (qop === 'auth' || qop === 'auth-int' || algorithm === 'MD5-sess') {
-            headerParams.push('nc=' + nonceCount);
+            nonceCount && nonceCountheaderParams.push('nc=' + nonceCount);
```
kamalaknn: `nonceCountheaderParams` -> `headerParams`

requires a unit test with `nonceCount`

PR #423, opened 2017-10-17
Make sure collection / folder level scripts are executed with request level scripts

kamalaknn:


kunagpal:
@kamalaknn Looks good overall, could you add tests for test scripts as well?

PR #427, opened 2017-10-20
 Allow entrypoint to filter by request

kamalaknn:


lib/runner/index.js
```diff
@@ -101,35 +104,39 @@ _.assign(Runner, {
...
-
-            group = idMatch || nameMatch; // Prioritise id based matches over name based ones.
+    extractRunnableItems: function (collection, entrypoint, callback) {
+        var items = [],
+            entryNode;
+
+        // if an entrypoint is specified, find the item/item group
+        // looks for first node with matching id or name
+        if (entrypoint) {
+            // @todo: fix this. the check by id should all finish before doing check by name
```
kamalaknn: Left the original comment here.

IMHO, we do not support this yet and I think we need hold off on this, now that the search goes through requests as well. Collections with names colliding with ids should be very rare and should not penalize performance of normal collections.

If we really want to support it, we have to make it a more performant search. It is common to have collections with thousands of requests.

kamalaknn: @kunagpal

kunagpal: In that case, `.oneDeep` will have to be adjusted to prioritize id matches over name matches.

test/unit/runner.test.js
```diff
@@ -128,16 +128,21 @@ describe('runner', function () {
     describe('extractRunnableItems', function () {
```
kunagpal: Leave existing tests alone, or move them to their own `describe` block for folders.

lib/runner/index.js
```diff
@@ -101,35 +104,39 @@ _.assign(Runner, {
...
+        else { entryNode = collection; }
+
+        // entry point lookup did not find anything
+        if (!entryNode) {
+            callback(null, items);
+            return;
+        }
+
+        // accumulate all items in order if entry point is a collection/folder
+        ItemGroup.isItemGroup(entryNode) ?
```
kunagpal: This can be made easier to read, the behaviour of prioritizing folder matches over requests should also be documented.

lib/runner/index.js
```diff
@@ -101,35 +104,39 @@ _.assign(Runner, {
...
-        // @todo: Possibly add mapItem to sdk.ItemGroup?
-        group && group.forEachItem && group.forEachItem(function (item) {
-            items.push(item);
-        });
+        // no entry point specified, start with the collection
+        else { entryNode = collection; }
+
+        // entry point lookup did not find anything
+        if (!entryNode) {
+            callback(null, items);
```
kunagpal: `return` should be on the same line as the `callback`.

test/unit/runner.test.js
```diff
@@ -152,20 +157,40 @@ describe('runner', function () {
             });
         });
 
-        it('should filter down to the provided groupId correctly', function (done) {
-            Runner.extractRunnableItems(collection, 'F1', function (err, runnableItems, entrypoint) {
+        it('should filter item correctly by name', function (done) {
```
kunagpal: Also add a test for folder-request name/id collisions.

lib/runner/index.js
```diff
@@ -101,35 +104,39 @@ _.assign(Runner, {
...
-        // accumulate all items in the collection/group in order
-        // @todo: Possibly add mapItem to sdk.ItemGroup?
-        group && group.forEachItem && group.forEachItem(function (item) {
-            items.push(item);
-        });
+        // no entry point specified, start with the collection
+        else { entryNode = collection; }
+
+        // entry point lookup did not find anything
+        if (!entryNode) {
```
kunagpal: This check should be in the `if` block on line 118.

lib/runner/index.js
kamalaknn: `.oneDeep` prioritizes `id` now see - https://github.com/postmanlabs/postman-collection/blob/develop/lib/collection/item-group.js#L210

lib/runner/index.js
kamalaknn: This is not prioritizing folder matches, a node can only be `Item` or `ItemGroup`, but not both.

test/unit/runner.test.js
kamalaknn: This PR does not handle name/id collisions.

test/unit/runner.test.js
kamalaknn: Moved `entrypoint` as folder and request into individual `describe` blocks.

lib/runner/index.js
kunagpal: In collections that have folders and requests with the same name, this will become a problem, but that will get resolved at the `oneDeep` call itself.

lib/runner/extract-runnable-items.js
```diff
@@ -0,0 +1,172 @@
...
+
+            /**
+             * Iteratee for looking up an element with given id on an item group.
+             *
+             * @param {ItemGroup} itemGroup
+             * @param {String} id
+             *
+             * @returns {Boolean|undefined}
+             */
+            idLookup = function (itemGroup, id) {
```
shamasis: anti-functional

lib/runner/extract-runnable-items.js
```diff
@@ -0,0 +1,172 @@
...
+     *
+     * @param {Collection} collection
+     * @param {String} id
+     *
+     * @returns {Item|ItemGroup|undefined}
+     *
+     * @todo: Possibly add oneDeepById to sdj.ItemGroup?
+     */
+    findItemOrGroupById = function (collection, id) {
+        var entry,
```
shamasis: itemOrGroup

lib/runner/extract-runnable-items.js
```diff
@@ -0,0 +1,172 @@
...
+     * @returns {Item|ItemGroup}
+     */
+    findItemOrGroupByPath = function (collection, path) {
+        var entryPath = path || [],
+            entryNode = collection,
+            i,
+            lookupLength = entryPath.length;
+
+        // go down the lookup path
+        for (i = 0; i < lookupLength && entryNode; i++) {
```
shamasis: for (i = 0; (i < lookupLength) && entryNode; i++)

shamasis: also check perf of


```
path = _.clone(path || []);

do {
    collection = collection && collections.items && collection.items.one(path.pop())
} while (collection && path.length);
```

lib/runner/extract-runnable-items.js
```diff
@@ -0,0 +1,172 @@
+var _ = require('lodash'),
```
shamasis: let's make a path lookup and idOrName lookup alone.

lib/runner/extract-runnable-items.js
shamasis:
```
entrypoint: {
   execute: '10e14f5f-880b-4ae5-a4c5-d96cc11b5882'
   lookup: 'path' // or 'idOrName',
   path: []
}
```

lib/runner/extract-runnable-items.js
kamalaknn: Checked the perf difference with `for` and `while with clone and pop`, the `while` options is 60% slower.

lib/runner/extract-runnable-items.js
kamalaknn: Done

kamalaknn: Couldn't get to a functional way to do this without shenanigans. Staying with synchronous flag based find now.

lib/runner/extract-runnable-items.js
```diff
@@ -0,0 +1,141 @@
...
+
+
+    /**
+     * Finds an item or group from a path. The path should be an array of ids from the parent chain.
+     *
+     * @param {Collection} collection
+     * @param {Object} options
+     * @param {Function} callback
+     */
+    lookupByPath = function (collection, options, callback) {
```
harryi3t: we could add docs for possible attributes for `options`
 - execute
 - path

kamalaknn: Yeah, good point

PR #430, opened 2017-10-23
Persist pm.variables across scripts

harryi3t:
~Blocked on https://github.com/postmanlabs/postman-sandbox/pull/223~

### DIFF
--- a/lib/runner/extensions/event.command.js
+++ b/lib/runner/extensions/event.command.js
@@ -259,6 +259,7 @@ module.exports = {
                             item: tempItem,
                             globals: payload.context.globals,
                             collectionVariables: payload.context.collectionVariables,
+                            _variables: payload.context._variables,
                             environment: payload.context.environment,
                             data: payload.context.data,
                             coords: cursor,
@@ -334,7 +335,9 @@ module.exports = {
                     result && result.response && (result.response = new sdk.Response(result.response));
 
                     // persist the pm.variables for the next script
-                    payload.context._variables = result._variables;
+                    payload.context._variables = new sdk.VariableScope(result._variables);
+                    // persist the pm.variables for the next request
+                    this.state._variables = new sdk.VariableScope(result._variables);
 
                     // now that this script is done executing, we trigger the event and move to the next script
                     this.triggers.script(err || null, cursor, result, script, event, item);
--- a/lib/runner/extensions/event.command.js
+++ b/lib/runner/extensions/event.command.js
@@ -8,8 +8,8 @@ var _ = require('lodash'),
     serialisedError = require('serialised-error'),
 
     ASSERTION_FAILURE = 'AssertionFailure',
-    SAFE_CONTEXT_VARIABLES = ['environment', 'globals', 'collectionVariables', 'cookies', 'data', 'request',
-        'response'],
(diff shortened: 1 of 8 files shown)


harryi3t:
@kamalaknn Tests are failing coz sandbox can't be used from any github branch.
Since it needs the `.cache/bootcode` to be present which is ignored in git but not in npm.
We need a released version for the tests to pass.

lib/runner/extensions/event.command.js
```diff
@@ -328,9 +329,16 @@ module.exports = {
                     // @todo - should this be handled by the sandbox?
                     result && result.environment && (result.environment = new sdk.VariableScope(result.environment));
                     result && result.globals && (result.globals = new sdk.VariableScope(result.globals));
+                    result && result.collectionVariables &&
+                        (result.collectionVariables = new sdk.VariableScope(result.collectionVariables));
```
kamalaknn: @shamasis we are bubbling `_variables` up with the results on the `test` callback. I'm guessing we are okay with this?

lib/runner/request-helpers-presend.js
```diff
@@ -17,6 +17,7 @@ module.exports = [
         // @todo - no need to sync vatiables when SDK starts supporting resolution from scope directly
         // @todo remove this hack of using empty variableList once it's fixed in sdk
         context.item = new sdk.Item(context.item.toObjectResolved({variables: new sdk.VariableList(null, [])}, [
+            context._variables.syncVariablesTo({}),
```
kamalaknn: @shamasis highlighting this change, `pm.variables.set` takes the highest precedence, higher than iteration data, the same as in `sandbox`

test/integration/inherited-entities/pm-variables.test.js
```diff
@@ -0,0 +1,307 @@
...
+        before(function (done) {
+            var clonedRunOptions = _.merge({}, runOptions, {
+                collection: {
+                    event: [
+                        {
+                            listen: 'prerequest',
+                            script: {
+                                exec: `
+                                // only run for 1st item
+                                if (pm.variables.get('key-1') !== 'modified-1') {
```
kamalaknn: @harryi3t can we try to use `unset` and avoid logic in the test fixtures to make it easy to comprehend?

harryi3t: ok will try it out

lib/runner/request-helpers-presend.js
```diff
@@ -14,9 +14,10 @@ module.exports = [
         if (!context.item) { return done(new Error('Nothing to resolve variables for.')); }
 
         // @todo - resolve variables in a more graceful way
+
         // @todo - no need to sync variables when SDK starts supporting resolution from scope directly
-        context.item = new sdk.Item(context.item.toObjectResolved(null, [context.data,
-            context.environment.syncVariablesTo({}), context.collectionVariables.syncVariablesTo({}),
+        context.item = new sdk.Item(context.item.toObjectResolved(null, [context._variables.syncVariablesTo({}),
```
shamasis: @kamalaknn - wondering if we should make this resolution as part of utils? It is used in two primary places?

```
resolveProperty = function (prop, context) ...
```

PR #440, opened 2017-11-02
Fix eslint issues in `integration-legacy` test directory

kamalaknn:


test/integration-legacy/abort-on-error-request-spec.test.js
```diff
@@ -10,71 +10,73 @@ describe('Option', function () {
             var errored = false,
                 runner = new runtime.Runner(),
                 rawCollection = {
-                    "variables": [],
-                    "info": {
-                        "name": "test",
-                        "_postman_id": "cd9e83b1-03dd-18ae-ff02-574414594a87",
-                        "description": "",
-                        "schema": "https://schema.getpostman.com/json/collection/v2.0.0/collection.json"
+                    'variables': [],
```
kunagpal: The quotes aren't actually necessary here. There's a flexible option for quote style, which lets you use both as needed.

PR #441, opened 2017-11-02
Fix eslint issues in `lib/requester` directory

kamalaknn:


lib/requester/browser/request.js
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/* eslint-disable */
```
kunagpal: ?

kamalaknn: It wouldn't be fixable without having to completely touch every line of the file. I'm not sure if we have such test coverage for the browser request flows.

kunagpal: Fine, to be done in a separate P.R then.

PR #443, opened 2017-11-03
Remove workaround for variable resolution `toObjectResolved`

kamalaknn:
The fix has been moved upstream


kunagpal:
@kamalaknn Builds are red.

PR #447, opened 2017-11-03
Add default value for global timeout

harryi3t:


lib/runner/index.js
```diff
@@ -3,6 +3,7 @@ var _ = require('lodash'),
     Run = require('./run'),
     extractRunnableItems = require('./extract-runnable-items').extractRunnableItems,
 
+    DEFAULT_GLOBAL_TIMEOUT = 3 * 60 * 1000, // 3 minutes
```
kamalaknn: @shamasis we are setting the default request timeout value.

PR #453, opened 2017-11-06
Sanitize timeout options to treat to 0 as Infinity

kamalaknn:


lib/runner/index.js
```diff
@@ -4,7 +4,35 @@ var _ = require('lodash'),
...
+
+    /**
+     * Sanitizes `timeout` `options` if any on the options object.
+     *
+     * @param {Object} [timeoutOptions]
+     *
+     * @returns {Object}
+     */
+    sanitizeTimeouts = function (timeoutOptions) {
+        if (!timeoutOptions) { return timeoutOptions; }
```
kunagpal: Shouldn't this be `return;`?

kamalaknn: Couple of checks that help us here,

1. `timeoutOptions` always has a default
2. We do not rely on return values on `sanitizeTimeouts`, the mutating behavior is relied on

That said you are right about falsy values, but just that the code flow never goes through.

PR #458, opened 2017-11-07
Update nsp to version 3.1.0

kunagpal:
Closes #455


shamasis:
Not updating to v3 since NSP dropped Node v4 support. We would get back in a quarter.

PR #460, opened 2017-11-07
Improvements to `timeout` value normalization

kamalaknn:
* Default timeouts are set if user provided timeout is missing/null/undefined
* To clear a timeout, 0 should be sent as timeout value
* Negative timeout values are treated as no timeout

### DIFF
--- a/lib/runner/index.js
+++ b/lib/runner/index.js
@@ -9,28 +9,6 @@ var _ = require('lodash'),
         global: 3 * 60 * 1000, // 3 minutes
         request: Infinity,
         script: Infinity
-    },
-
-    /**
-     * Sanitizes run options. Only handles timeout for now.
-     *
-     * @param {Object} options
-     *
-     * @returns {Object}
-     */
-    sanitizeRunOptions = function (options) {
-        !options.timeout && (options.timeout = {});
-
-        // start timeout sanitization
-        _.mergeWith(options.timeout, defaultTimeouts, function (userTimeout, defaultTimeout) {
-            // non numbers, Infinity and missing values are set to default
-            if (!_.isFinite(userTimeout)) { return defaultTimeout; }
-
-            // 0 and negative numbers are set to Infinity, which only leaves positive numbers
-            return userTimeout > 0 ? userTimeout : Infinity;
-        });
-
-        return options;
     };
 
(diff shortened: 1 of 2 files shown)


lib/runner/index.js
```diff
@@ -3,35 +3,34 @@ var _ = require('lodash'),
...
 
-        _.forEach(timeoutOptions, function (value, key) {
-            timeoutOptions[key] = sanitizeTimeout(value);
+        // start timeout sanitization
+        _.mergeWith(options.timeout, defaultTimeouts, function (userTimeout, defaultTimeout) {
+            // non numbers, Infinity and missing values are set to default
+            if (!_.isFinite(userTimeout)) { return defaultTimeout; }
+
+            // 0 and negative numbers are set to Infinity, which only leaves positive numbers
+            return userTimeout > 0 ? userTimeout : Infinity;
```
harryi3t: @shamasis @kamalaknn Just wanted to bring attention to this
Ideally, this should not work for `vm`. 
It working coz Infinity is being converted to `null` during dispatching.

test/unit/runner.test.js
```diff
@@ -154,42 +157,43 @@ describe('runner', function () {
                     var runner = new Runner({
                         run: {
                             timeout: {
-                                global: runnerTimeout,
```
harryi3t: let's not remove this, please :)

lib/runner/index.js
```diff
@@ -75,17 +74,13 @@ _.assign(Runner.prototype, {
      */
     run: function (collection, options, callback) {
         var self = this,
-            runOptions = _.merge({timeout: {
-                global: DEFAULT_GLOBAL_TIMEOUT,
-                script: 0,
-                request: 0
-            }}, _.omit(options, ['environment', 'globals', 'data']), this.options.run);
+            runOptions = _.merge(_.omit(options, ['environment', 'globals', 'data']), this.options.run) || {};
```
shamasis: merge this code and the `sanitizeRunOptions` to one `this.config(options)`. 
Move `sanitizeRunOptions` as `run.prototype.config` or some other suitable name.

Parsing self's options wrt to its own options seems to me as a task for the prototype.

PR #470, opened 2017-11-13
Maintain auth state and document the state requirements for every auth

harryi3t:


lib/authorizer/digest.js
```diff
@@ -90,6 +90,10 @@ function _getDigestAuthHeader (headers) {
 
 
 /**
+ * `realm` and `nonce` should be stored and reused for Digest auth. If not found or if the the server responds with 401,
+ * a second request should be sent using the values sent by the server.
```
harryi3t: This should be `If not found, a second request....` 
Since we don't override the user provided values

lib/authorizer/oauth2.js
```diff
@@ -13,6 +13,8 @@ var _ = require('lodash'),
     ];
 
 /**
+ * OAuth-2 does not require to maintain it's state.
```
shamasis: Can you drop all the “does not require”?! And only add the one where it’s required and with a bit more explanation. :-p

harryi3t: sure :)

kamalaknn:
@shamasis This would be okay to build on top of `Use Auth definition for storing state` until we have a proper state.

My concerns for not taking this approach are

1. Some of the properties here do not make sense on Auth definition. They only make sense in the intermediate cycles.
2. The only way to persist auth after a collection run would be to save the `Auth definition`, which might cause unwanted side effects.
3. The responsibility of keeping the original auth in sync is scattered and in no place. Can we put this in the `AuthInterface` if we are going down this path?

lib/runner/request-helpers-postsend.js
```diff
@@ -16,6 +16,7 @@ module.exports = [
         if (!(context.auth && context.auth.type)) { return done(); }
 
         var auth = context.auth,
+            originalAuthParams = context.originalItem.getAuth().parameters(),
```
kamalaknn: Make sure that `originalItem` has an `auth`. Eventhough, it will always be if copied from `item`, but we don't want to rely on the assumption.

PR #471, opened 2017-11-13
Add support for custom dns lookup

harryi3t:
This only adds support for resolving hostnames by using an object.
How that object is created is currently left to discuss (Should app do it or runtime?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,22 @@
 * More useful error messages for assertion failures in legacy `tests`
 * Digest auth does not attempt retries for invalid credentials/configuration. It will continue to retry for missing configuration.
 * Auth will maintain it's state across a collection run. Digest auth no longer needs to send 2 requests for every digest auth protected item in a collection
+* Added support for custom DNS lookup
+* [Breaking] `restrictedAddresses` option is now moved to `network` option
+    - In v7.0.0
+        ```js
+        runner.run(collection, {
+            restrictedAddresses: {'x.x.x.x': true}
+        }});
+        ```
+    - In Unreleased
+        ```js
+        runner.run(collection, {
+            network: {
+                restrictedAddresses: {'x.x.x.x': true}
+            }
+        }});
+        ```
 
 ### v7.0.1 (November 8, 2017)
 * :bug: Fixed a bug where the assertions for legacy `tests` failures did not have an `error` object.
--- a/README.md
+++ b/README.md
@@ -84,8 +84,17 @@ runner.run(collection, {
         // Enable sending of bodies with GET requests (only supported on Node, ignored in the browser)
         sendBodyWithGetRequests: true,
 
-        // Allows restricting IP/host in requests
-        restrictedAddresses: {'192.168.1.1': true}
+        // network related options
+        network: {
+            hosts: { // hosts configuration for dns lookup
+                type: 'hostIpMap',
+                hostIpMap: {
+                    'domain.com': '127.0.0.1',
+                    'ipv6-domain.com': '::1',
+                }
+            },
+            restrictedAddresses: {'192.168.1.1': true} // Allows restricting IP/host in requests
+        }
     },
 
     // A ProxyConfigList, from the SDK
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ runner.run(collection, {
 
         // network related options
         network: {
-            hosts: { // hosts configuration for dns lookup
+            hostLookup: { // hosts file configuration for dns lookup
                 type: 'hostIpMap',
                 hostIpMap: {
                     'domain.com': '127.0.0.1',
--- a/lib/authorizer/ntlm.js
+++ b/lib/authorizer/ntlm.js
@@ -107,6 +107,7 @@ module.exports = {
         }
 
         if (state === STATES.INITIALIZED) {
+            console.log('response.headers', response.headers);
 
             // Nothing to do if the server does not ask us for auth in the first place.
             if (!(response.headers.has(WWW_AUTHENTICATE, NTLM) ||
--- a/lib/authorizer/ntlm.js
+++ b/lib/authorizer/ntlm.js
@@ -107,8 +107,6 @@ module.exports = {
         }
 
         if (state === STATES.INITIALIZED) {
-            console.log('response.headers', response.headers);
-
             // Nothing to do if the server does not ask us for auth in the first place.
             if (!(response.headers.has(WWW_AUTHENTICATE, NTLM) ||
                   response.headers.has(WWW_AUTHENTICATE, NEGOTIATE))) {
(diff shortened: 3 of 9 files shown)


kunagpal:
Shouldn't the behaviour be to add a flag to use `/etc/hosts` or a Windows equivalent, true by default?
This will effectively require users to actively supply the set of preferred resolutions.

lib/requester/requester-pool.js
```diff
@@ -14,7 +14,8 @@ RequesterPool = function (options) {
         strictSSL: _.get(options, 'requester.strictSSL'),
         sendBodyWithGetRequests: _.get(options, 'requester.sendBodyWithGetRequests', false),
         followRedirects: _.get(options, 'requester.followRedirects', true),
-        restrictedAddresses: _.get(options, 'restrictedAddresses')
+        restrictedAddresses: _.get(options, 'restrictedAddresses'),
```
shamasis: Now we have two of these IP like confit sent to requester and is very specific network stuff. They might need to be placed somewhere or inside its own namespace in options.

lib/requester/util.js
```diff
@@ -68,14 +68,23 @@ var _ = require('lodash'),
...
     lookup = function (lookupOptions, hostname, options, callback) {
-        var self = this;
+        var self = this,
+            lowercaseHost = hostname.toLowerCase(),
+            resolvedAddr;
+
+        if (lowercaseHost !== LOCALHOST) {
+            if (lookupOptions.hostIpMap && (resolvedAddr = lookupOptions.hostIpMap[lowercaseHost])) {
+                typeof resolvedAddr === 'string' && (resolvedAddr = {ip: resolvedAddr, family: 4});
+                return callback(null, resolvedAddr.ip, resolvedAddr.family || 4);
```
shamasis: We blindly trust this to be an IP address? What if it is not?

harryi3t: yes, will add checks for it.
Also, I just realized that this is bypassing the check for `restrictedAddresses`. Will fix that too.

lib/requester/requester-pool.js
kamalaknn: This can be done and makes sense, but doing it should be a breaking change and we've just released `7.0`. 

Other option would be to support both `options.restrictedAddresses` and `options.network.{restrictedAddresses,hostIpMap}`.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * More useful error messages for assertion failures in legacy `tests`
 * Digest auth does not attempt retries for invalid credentials/configuration. It will continue to retry for missing configuration.
 * Auth will maintain it's state across a collection run. Digest auth no longer needs to send 2 requests for every digest auth protected item in a collection
+* Added support for custom DNS lookup
```
kamalaknn: Changelog needs to have an entry for the breaking change in `restrictedAddresses`

kamalaknn: Also `README` needs to be updated.

lib/requester/util.js
```diff
@@ -67,18 +71,40 @@ var _ = require('lodash'),
      *
      * @param lookupOptions
      * @param lookupOptions.port
-     * @param lookupOptions.restrictedAddresses
+     * @param lookupOptions.network
+     * @param lookupOptions.network.restrictedAddresses
+     * @param lookupOptions.network.hosts
+     * @param lookupOptions.network.hosts.type
```
kamalaknn: `@param lookupOptions.network.hosts.hostIpMap`

lib/requester/util.js
```diff
@@ -67,18 +71,40 @@ var _ = require('lodash'),
...
+            resolvedFamily = 4,
+            resolvedAddr;
+
+        hostsOptions && hostsOptions.type === HOSTS_TYPE.HOST_IP_MAP &&
+            (hostIpMap = hostsOptions[HOSTS_TYPE.HOST_IP_MAP]) && (resolvedAddr = hostIpMap[lowercaseHost]);
+
+        if (resolvedAddr) {
+            _.includes(resolvedAddr, COLON) && (resolvedFamily = 6);
+
+            // returning error synchronously causes uncaught error
```
kamalaknn: Can you add
`// because listeners are not attached to error events on socket yet`

lib/requester/util.js
```diff
@@ -67,18 +71,40 @@ var _ = require('lodash'),
...
+            return setImmediate(function () {
+                callback(self.isAddressRestricted(resolvedAddr, networkOpts) ?
```
kamalaknn: Could you extract the lookup function so that `dns.lookup` and `hostIpMap` lookup are passed with the same callback and `isAddressRestricted` is handled in there?

This looks prone to errors.

lib/requester/util.js
```diff
@@ -155,13 +182,13 @@ module.exports = {
         self.ensureHeaderExists(options.headers, 'Host', request.url.getRemote());
 
         // override DNS lookup
-        if (defaultOpts.restrictedAddresses || hostname.toLowerCase() === LOCALHOST) {
+        if (!_.isEmpty(networkOptions) || hostname.toLowerCase() === LOCALHOST) {
```
kamalaknn: This is critical path no `lodash` here please.

CHANGELOG.md
```diff
@@ -5,6 +5,22 @@
...
+    - In v7.0.0
+        ```js
+        runner.run(collection, {
+            restrictedAddresses: {'x.x.x.x': true}
+        }});
+        ```
+    - In Unreleased
+        ```js
+        runner.run(collection, {
+            network: {
```
kamalaknn: @shamasis `network` sounds good as an option?

lib/requester/util.js
```diff
@@ -402,12 +449,13 @@ module.exports = {
...
+     * @param {Object} networkOptions
+     * @param {Array<String>} networkOptions.restrictedAddresses
      *
      * @returns {Boolean}
      */
-    isAddressRestricted: function (host, options) {
-        return options.restrictedAddresses && options.restrictedAddresses[(host && host.toLocaleLowerCase())];
+    isAddressRestricted: function (host, networkOptions) {
+        return networkOptions.restrictedAddresses &&
+            networkOptions.restrictedAddresses[(host && host.toLocaleLowerCase())];
```
kamalaknn: Should be `toLowerCase`? My mistake.

test/integration/sanity/dns-lookup.test.js
```diff
@@ -0,0 +1,51 @@
+var dns = require('dns');
+
+describe('DNS lookup', function() {
+    var testrun;
+
+    before(function(done) {
+        var self = this;
+
+        // Not hard-coding since this can change
```
kamalaknn: 👍

README.md
```diff
@@ -84,8 +84,17 @@ runner.run(collection, {
         // Enable sending of bodies with GET requests (only supported on Node, ignored in the browser)
         sendBodyWithGetRequests: true,
 
-        // Allows restricting IP/host in requests
-        restrictedAddresses: {'192.168.1.1': true}
+        // network related options
+        network: {
+            hosts: { // hosts configuration for dns lookup
```
kamalaknn: shouldn't this be something related to `lookup`?

harryi3t: renaming to `hostLookup`

lib/authorizer/ntlm.js
```diff
@@ -107,6 +107,7 @@ module.exports = {
         }
 
         if (state === STATES.INITIALIZED) {
+            console.log('response.headers', response.headers);
```
kamalaknn: ??

PR #475, opened 2017-11-14
Moved variable resolution out of `http-request` extension

kamalaknn:
* `http-request` extension is now only responsible for request send related responsibilities
* Variable resolution is now the responsibility of `request` extension, there is no need to send all variable types to `http-request` from `event` extension and `replay-controller`.
* Context creation extracted out to be used by anyone scheduling a request with `http-request` extension

### DIFF
--- a/lib/runner/create-item-context.js
+++ b/lib/runner/create-item-context.js
@@ -1,0 +1,50 @@
+var _ = require('lodash'),
+    sdk = require('postman-collection'),
+
+    SAFE_CONTEXT_PROPERTIES = ['replayState', 'coords'],
+
+    /**
+     * Creates a context object to be used with `http-request.command` extension.
+     *
+     * @param {Object} payload
+     * @param {Item} payload.item
+     * @param {Object} [payload.coords]
+     * @param {Object} [defaults]
+     * @param {Object} [defaults.replayState]
+     * @param {Object} [defaults.coords]
+     *
+     * @returns {Object}
+     */
+    createItemContext = function (payload, defaults) {
+        // extract properties from defaults that can/should be reused in new context
+        var context = defaults ? _.pick(defaults, SAFE_CONTEXT_PROPERTIES) : {};
+
+        // set cursor to context
+        !context.coords && (context.coords = payload.coords);
+
+        // save original item for reference
+        context.originalItem = payload.item;
+
+        // we clone item from the payload, so that we can make any changes we need there, without mutating the
+        // collection
+        context.item = new sdk.Item(payload.item.toJSON());
+
+        // get a reference to the Auth instance from the item, so changes are synced back
+        context.auth = context.originalItem.getAuth();
+
+        /**
+         * @type {Object}
+         * @property {Object} coords - current cursor
+         * @property {Item} originalItem - reference to the item in the collection
+         * @property {Item} item -  Holds a copy of the item given in the payload, so that it can be manipulated
+         * as necessary
+         * @property {RequestAuthBase|undefined} auth - If present, is the instance of Auth in the collection, which
+         * is changed as necessary using intermediate requests, etc.
+         * @property {ReplayState} replayState - has context on number of replays(if any) for this request
+         */
+        return context;
+    };
+
+module.exports = {
+    createItemContext: createItemContext
+};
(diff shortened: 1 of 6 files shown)


lib/runner/create-item-context.js
```diff
@@ -0,0 +1,50 @@
...
+
+        // set cursor to context
+        !context.coords && (context.coords = payload.coords);
+
+        // save original item for reference
+        context.originalItem = payload.item;
+
+        // we clone item from the payload, so that we can make any changes we need there, without mutating the
+        // collection
+        context.item = new sdk.Item(payload.item.toJSON());
```
kunagpal: Thinking out loud: There should be a clone method in the SDK 🤔

lib/runner/create-item-context.js
```diff
@@ -0,0 +1,50 @@
...
+
+        // get a reference to the Auth instance from the item, so changes are synced back
+        context.auth = context.originalItem.getAuth();
+
+        /**
```
kunagpal: ?

lib/runner/create-item-context.js
```diff
@@ -0,0 +1,50 @@
...
+         * as necessary
+         * @property {RequestAuthBase|undefined} auth - If present, is the instance of Auth in the collection, which
+         * is changed as necessary using intermediate requests, etc.
+         * @property {ReplayState} replayState - has context on number of replays(if any) for this request
+         */
+        return context;
+    };
+
+module.exports = {
+    createItemContext: createItemContext
```
kunagpal: The function could be defined here directly.

lib/runner/extensions/event.command.js
```diff
@@ -7,6 +7,8 @@ var _ = require('lodash'),
     sandbox = require('postman-sandbox'),
     serialisedError = require('serialised-error'),
 
+    createItemContext = require('../create-item-context').createItemContext,
```
kunagpal: This looks odd.

lib/runner/extensions/request.command.js
```diff
@@ -1,4 +1,42 @@
...
+        // @todo - no need to sync variables when SDK starts supporting resolution from scope directly
+        context.item = new sdk.Item(context.item.toObjectResolved(null, [payload._variables.syncVariablesTo({}),
+            payload.data, payload.environment.syncVariablesTo({}), payload.collectionVariables.syncVariablesTo({}),
+            payload.globals.syncVariablesTo({})], {ignoreOwnVariables: true}));
+
+        var item = context.item,
+            auth = context.auth;
+
+        // Re-parse the URL, because variables have been resolved now, and things might be moved around
+        item.request.url = new (sdk.Url)(item.request.url.toString());
```
kunagpal: This feels like a workaround 😓.

lib/runner/extensions/request.command.js
```diff
@@ -1,4 +1,42 @@
...
+
+        // resolve variables in auth
+        auth && (context.auth = new sdk.RequestAuth(auth.toObjectResolved(null, [payload._variables.syncVariablesTo({}),
```
kunagpal: This and the section above could be extracted into a common method.

lib/runner/replay-controller.js
```diff
@@ -43,13 +44,16 @@ _.assign(ReplayController.prototype, /** @lends ReplayController.prototype */{
         context.replayState = this.getReplayState();
 
         // construct payload for request
-        var payload = _.merge({}, desiredPayload, context, {
+        var payload = _.merge({}, desiredPayload, {
```
kunagpal:
```js
_.defaults({
  item: item,
  abortOnError: true
}, desiredPayload);
```

lib/runner/extensions/request.command.js
kamalaknn: We'll be adding a proper controller for resolving variables. Otherwise, it wouldn't give us much of organization.

lib/runner/create-item-context.js
kamalaknn: Maybe 🤔

lib/runner/extensions/request.command.js
kamalaknn: Will take a look at it.

PR #479, opened 2017-11-16
Added `scriptId` and `eventId` to track logs, errors to scripts

kamalaknn:


lib/runner/extensions/event.command.js
```diff
@@ -215,75 +215,86 @@ module.exports = {
...
+                    // create copy of cursor so we don't leak script ids outside `event.command`
+                    // and across scripts
+                    scriptCursor = _.clone(cursor);
 
                 // store the execution id in script
                 script._lastExecutionId = executionId; // please don't use it anywhere else!
 
+                // if we can find an id on script or event we add them to the cursor
+                // so logs and errors can be traced back to the script they came from
+                event.id && (scriptCursor.eventId = event.id);
```
kamalaknn: @shamasis `event.id` is not made required in `postman-collection`, also in not included in the schema. Should I?

kunagpal: Let's use `_postman_id` instead.

kamalaknn: Umm, I think we shouldn't. This is supposed to help consumers track any logs or errors back to the script and event that triggered them. I am not sure if internal ids will be useful with that.

shamasis: Add to sdk. No need to add to Schrems or transformer.

lib/runner/extensions/event.command.js
```diff
@@ -217,69 +217,80 @@ module.exports = {
...
-                            this.host.dispatch(
-                                EXECUTION_RESPONSE_EVENT_BASE + id,
-                                requestId,
-                                null,
-                                result && result.response
-                            );
-                        }).catch(function (err) {
-                            this.host.dispatch(EXECUTION_RESPONSE_EVENT_BASE + id, requestId, err);
-                        });
+                this.host.on(EXECUTION_REQUEST_EVENT_BASE + executionId,
```
shamasis: Wondering why’s this shown as a diff.

shamasis: Was this crossing max line length lint rule?

kamalaknn: The diff on this line is

```diff
-this.host.on(EXECUTION_REQUEST_EVENT_BASE + executionId, function (cursor, id, requestId, request) {
+this.host.on(EXECUTION_REQUEST_EVENT_BASE + executionId, function (scriptCursor, id, requestId, request) {
```

This exceded max line length, and the newline on the function parameter needed one extra indent level, so that is the rest of the diff.

kamalaknn:
@shamasis agreed that `cursor` is becoming a filmsy construct now. We'll need to design it through. If it doesn't feel right to have sub-item level pointers, we need to create one that can contain these. More and more features are required to be built around sub-item level pointers.

lib/runner/extensions/event.command.js
harryi3t: @kamalaknn Maybe we also update the README's assertion callback doc. We are listing all the properties of cursor there.

PR #490, opened 2017-11-27
Do not convert variable lists to POJOs for resolution

kamalaknn:
Converting variable list POJO leads to data loss on meta information like `disabled` and `system`. This prevents SDK from adding functionality based on the `meta` information.


kamalaknn:
We have tests to cover successful variable resolution at each level and the precedence in case of duplicates. New tests should not be necessary.

lib/runner/extensions/request.command.js
```diff
@@ -22,9 +22,16 @@ var _ = require('lodash'),
         // @todo - resolve variables in a more graceful way
 
         // @todo - no need to sync variables when SDK starts supporting resolution from scope directly
-        context.item = new sdk.Item(context.item.toObjectResolved(null, [payload._variables.syncVariablesTo({}),
-            payload.data, payload.environment.syncVariablesTo({}), payload.collectionVariables.syncVariablesTo({}),
-            payload.globals.syncVariablesTo({})], {ignoreOwnVariables: true}));
+        context.item = new sdk.Item(context.item.toObjectResolved(null, [
```
kunagpal: Code repetition 😓, looks good otherwise.

PR #498, opened 2017-12-04
Remove duplicate headers and query params in OAuth 2.0 before signing

kamalaknn:
Postman App indicates that duplicate headers will be removed. This behaviour of adding without removing creates two `Authorization` headers.


kunagpal:
@kamalaknn Looks like the synchronous script timeout breach tests are flaky 😞

PR #500, opened 2017-12-07
Fixed entrypoint error message, added test

deepakpathania:
- Fixes entrypoint error message issues in newman discussed [here](https://github.com/postmanlabs/newman/pull/1326).

### DIFF
--- a/lib/runner/index.js
+++ b/lib/runner/index.js
@@ -79,8 +79,7 @@ _.assign(Runner.prototype, {
      */
     run: function (collection, options, callback) {
         var self = this,
-            runOptions = this.prepareRunConfig(options),
-            errorPath;
+            runOptions = this.prepareRunConfig(options);
 
         callback = backpack.normalise(callback);
         !_.isObject(options) && (options = {});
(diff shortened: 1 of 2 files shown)


lib/runner/index.js
```diff
@@ -89,10 +90,12 @@ _.assign(Runner.prototype, {
 
             // Bail out only if: abortOnError is set and the returned entrypoint is invalid
             if (options.abortOnError && !entrypoint) {
-                return callback(new Error(`Unable to find a folder or request: ${options.entrypoint}`));
+                (_.isObject(options.entrypoint) && options.entrypoint.execute) ?
+                    (errorPath = options.entrypoint.execute) :
+                    (errorPath = options.entrypoint);
+                return callback(new Error(`Unable to find a folder or request: ${errorPath}`));
```
kunagpal:
```js
${_.get(options, 'entrypoint.execute', options.entrypoint);}
```

kamalaknn: @kunagpal `options.entrypoint` is not polymorphic, i.e. `entrypoint: folderName` is not allowed.  `entrypoint` should always be an object. 

If `entrypoint` is passed in an invalid form, runtime might not parse it or use it, but the error parses it, which might be confusing.

I would suggest `Unable to find folder or request: ${_.get(options, 'entrypoint.execute')}` so that the error message reads `Unable to find folder or request: undefined` when used in unsupported format.

deepakpathania: @kamalaknn There is actually a unit [test](https://github.com/postmanlabs/postman-runtime/blob/develop/test/unit/runner.test.js#L38) in runtime that checks for this behavior and that specifies the entrypoint string in the error message in case `abortOnError` is true with an invalid entrypoint string. I thought this was the intended behavior.

kamalaknn: @deepakpathania The test on runtime is a false positive. That also needs to be updted to test for `entrypoint: { execute: 'random' }`.

entrypoint as a string is not to be supported. Could you make sure integration/unit tests in runtime are updated to reflect that?

kunagpal:
@deepakpathania A quick note: the script timeout tests here are actually flaky. 🤔

PR #616, opened 2018-07-19
Avoid forwarding certificate lookup error

codenirvana:
FIxes postmanlabs/postman-app-support#3251

### DIFF
--- a/lib/runner/request-helpers-presend.js
+++ b/lib/runner/request-helpers-presend.js
@@ -306,7 +306,6 @@ module.exports = [
             if (err) {
                 // Swallow the error after triggering a warning message for the user.
                 run.triggers.console(context.coords, 'warn', 'certificate load error: ' + (err.message || err));
-                return done();
             }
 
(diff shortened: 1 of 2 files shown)


lib/runner/request-helpers-presend.js
```diff
@@ -306,7 +306,6 @@ module.exports = [
             if (err) {
                 // Swallow the error after triggering a warning message for the user.
                 run.triggers.console(context.coords, 'warn', 'certificate load error: ' + (err.message || err));
-                return done();
```
shamasis: removing done here does not solve it. Async would still bail out from mapValues since fileResolver iterator function returns an error.

The `return done();` should be retained and instead L:304 needs to be changed to:

```
function (value, key, next) {
    fileResolver.readFile(value, function (err, data) {
        err && run.triggers.console(context.coords, 'warn', 'certificate load error: ' + (err.message || err));
        next(null, data);
    });
}

shamasis: Also, add failing test case.

PR #632, opened 2018-09-03
Added tests for custom HTTP method

codenirvana:


shamasis:
@codenirvana @kunagpal missing test for blank/missing/non-string request methods.

PR #636, opened 2018-09-07
 Handle `disabled` request body

codenirvana:


lib/runner/request-helpers-presend.js
```diff
@@ -159,6 +159,10 @@ module.exports = [
 
         if (!request) { return done(new Error('No request to send.')); }
 
+        // if body is disabled than skip loading files.
+        // @todo this may cause problem if body is enabled/disabled programmatically from pre-request script.
```
kunagpal: Changes made to `pm.request` will not have any effect on the actual being sent (for now)

test/unit/requester-core.test.js
```diff
@@ -287,6 +287,20 @@ describe('requester util', function () {
...
+                url: 'postman-echo.com/post',
+                method: 'POST',
+                body: {
+                    disabled: true,
+                    mode: 'raw',
+                    raw: '{"beta":"bar"}'
+                }
+            });
+
+            expect(requesterCore.getRequestBody(request)).to.be.undefined;
```
kunagpal: Is this also the behaviour when no body exists?

test/integration/file-uploads/request-body.test.js
```diff
@@ -121,6 +121,52 @@ describe('file upload in request body', function () {
         });
     });
 
+    describe('with disabled body', function () {
```
kunagpal: Also add a test for the body being sent with `disabled: false`

test/unit/requester-core.test.js
codenirvana: Yes

PR #643, opened 2018-09-20
 Added tests for protocolProfileBehavior using HEAD & Custom request method

codenirvana:
Verifies postmanlabs/postman-request/pull/23


test/integration/request-body/protocolProfileBehavior.test.js
```diff
@@ -4,6 +4,7 @@ var net = require('net'),
 describe('protocolProfileBehavior', function () {
     var server,
```
kunagpal: Add an `afterEach` block inside this describe block to set `rawRequest` to `null`.

PR #647, opened 2018-10-12
Migrate System Tests to Chai

bharathr145:
- Use chai expect assertions
- Add global expect for System test

### DIFF
--- a/test/system/repository.test.js
+++ b/test/system/repository.test.js
@@ -69,7 +69,9 @@ describe('project repository', function () {
 
             // Hawk library v6.0.2+ uses ES2017 and is not compatible with Node <8.
             it('hawk version', function () {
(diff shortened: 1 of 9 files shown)


test/system/repository.test.js
```diff
@@ -3,78 +3,76 @@
...
-            it('must exist', function () {
-                expect(json.dependencies).to.be.a('object');
+            it('should exist', function () {
+                expect(json.dependencies).to.be.an('object');
             });
 
             // Hawk library v6.0.2+ uses ES2017 and is not compatible with Node <8.
             it('hawk version', function () {
-                expect(json.dependencies.hawk).to.be('6.0.2');
+                expect(json.dependencies.hawk).to.equal('6.0.2');
```
kunagpal: Use the deep assertion helper instead.

PR #713, opened 2019-01-18
Exposed runtime version and dependencies tree via API

vikiCoder:


lib/version.js
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Generate dependency tree of runtime.
+ *
+ * @param {Object} [tree={}] - Object to store dependencies
+ * @param {String} [moduleName] - Module name
+ * @returns {Object} - Nested object of dependencies
+ *
+ * @note Don't use any arguments in this function, those are for internal use only.
+ */
+dependenciesTree = function (tree, moduleName) {
```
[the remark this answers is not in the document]
vikiCoder: Updated API.

lib/version.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+    else {
+        // runtime's dependency package
+        packageJson = require('../node_modules/' + moduleName + '/package.json');
+    }
+
+    // normal and dev dependencies combined
+    dependencies = Object.assign({}, packageJson.dependencies, packageJson.devDependencies);
+
+    for (key in dependencies) {
```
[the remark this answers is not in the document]
vikiCoder: changed outer loop. Now it iterates on PACKAGES_TO_CONSIDER array instead of all dependencies.

lib/version.js
```diff
@@ -0,0 +1,108 @@
...
+        'postman-sandbox',
+        'uniscope',
+        'uvm'
+    ],
+
+    /**
+     * Generate dependency tree of runtime.
+     *
+     * @param {String} [dependencyName] - Module name
+     * @param {Boolean} [findDeepDependencies=true] - Do depth dependencies traversal or stop at root
```
codenirvana: Just `deepDependencies` or `oneDeep` 🤔

lib/version.js
```diff
@@ -0,0 +1,108 @@
...
+        'chai-postman',
+        'postman-collection',
+        'postman-request',
+        'postman-sandbox',
+        'uniscope',
+        'uvm'
+    ],
+
+    /**
+     * Generate dependency tree of runtime.
```
codenirvana: Need to update the description to indicate it returns dependencies versions as well.

lib/version.js
```diff
@@ -0,0 +1,108 @@
...
+     *
+     * @param {String} [dependencyName] - Module name
+     * @param {Boolean} [findDeepDependencies=true] - Do depth dependencies traversal or stop at root
+     * @param {Object} [dependencyTree={}] - Object to store dependencies
+     * @returns {Object}
+     *
+     * @note Don't use tree argument in this function, that is for internal use only.
+     *
+     * @example <caption>Returned Object Structure</caption>
+     * {
```
codenirvana: Missing how to get this result.

lib/version.js
```diff
@@ -0,0 +1,108 @@
...
+     *     }
+     * }
+     */
+    dependenciesTree = function (dependencyName, findDeepDependencies, dependencyTree) {
+        !dependencyTree && (dependencyTree = {});
+        (findDeepDependencies === undefined) && (findDeepDependencies = true);
+
+        var version,
+            // runtime's package.json
+            packageJson = require('../package.json'),
```
codenirvana: This can be moved to top as constant since this statement is not required for every recursive call.

lib/version.js
```diff
@@ -0,0 +1,108 @@
...
+            key;
+
+        if (dependencyName) {
+            if (
```
codenirvana: Avoid nested IFs

test/unit/version.test.js
```diff
@@ -0,0 +1,49 @@
...
+                        uniscope: {
+                            version: sandboxPackage.devDependencies.uniscope
+                        }
+                    }
+                }
+            }
+        };
+
+    it('should export correct runtime version', function () {
+        expect(runtime.version().version).to.eql(version.version);
```
codenirvana: Also, add a test with dependency argument.

lib/version.js
```diff
@@ -0,0 +1,108 @@
+var /**
```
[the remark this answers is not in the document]
vikiCoder: No need after new changes.

lib/version.js
vikiCoder: I chose deepDependencies

lib/version.js
vikiCoder: How does it look now?

lib/version.js
vikiCoder: Added function call.

lib/version.js
vikiCoder: Done.

lib/version.js
```diff
@@ -0,0 +1,115 @@
...
+     *         },
+     *         'postman-request': {
+     *             version: '2.88.1-postman.5'
+     *         }
+     *     }
+     * }
+     */
+    getVersionData = function (moduleName, deepDependencies, moduleData) {
+        !moduleData && (moduleData = {});
+        (deepDependencies === undefined) && (deepDependencies = true);
```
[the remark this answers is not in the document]
vikiCoder: If you are referring to the function getVersionData, , it generates and returns the dependency object as given in the comment above.

Or if you are referring to the first two lines of the function, those are to set the default values of function arguments. 

- The argument moduleData will contain an object with version of dependency given in the moduleName argument(if function is called recursively, empty object otherwise).

- The argument deepDependencies is a boolean (true by default) indecating weather to include recursive dependency in generated object or not.

lib/version.js
```diff
@@ -0,0 +1,115 @@
+var RUNTIME_PACKAGE = require('../package.json'),
```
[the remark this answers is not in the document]
vikiCoder: As I understand, Lodash functions are slow compared to a simple approach. Besides, there is not much need to use them here anyway.

lib/version.js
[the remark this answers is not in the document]
kunagpal: Add a comment explaining that this is used for the recursive call, so that similar confusions can be avoided in the future.

PR #719, opened 2019-01-25
Added SSL server to separate module in tests

vikiCoder:
Removed duplicate code for local SSL server from test files and added it in server.js module.

### DIFF
--- a/test/fixtures/server.js
+++ b/test/fixtures/server.js
@@ -93,11 +93,7 @@ function createSSLServer (opts) {
             'ca': path.join(certDataPath, 'ca-crt.pem')
         };
 
-    if (opts) {
-        for (i in opts) {
-            options[i] = opts[i];
-        }
-    }
+    options = Object.assign(options, opts);
 
     for (i in options) {
         if (i !== 'requestCert' && i !== 'rejectUnauthorized' && i !== 'ciphers') {
--- a/test/fixtures/server.js
+++ b/test/fixtures/server.js
@@ -71,40 +71,42 @@ function createRawEchoServer () {
 }
 
 /**
- * Simple SSL server for tests. Send
- *  - 'authorized' to authorized clients.
- *  - 'unauthorized' to unauthorized clients.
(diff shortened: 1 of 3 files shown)


test/fixtures/server.js
```diff
@@ -67,6 +70,49 @@ function createRawEchoServer () {
...
+ *  - 'authorized' to authorized clients.
+ *  - 'unauthorized' to unauthorized clients.
+ *
+ * @param {Boolean} [requestCert=true] - If true, request a certificate from clients that connect
+ *
+ * @example
+ * var s = createSslServer();
+ * s.listen(3000, 'localhost');
+ */
+function createSslServer (requestCert) {
```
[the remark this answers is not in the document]
codenirvana: Accept `key`, `cert` and `ca` from argument just like `requestCert`.
Also, this can be extended to multiple options in the future so it's better to accept the entire options as an argument and still have defaults as a fallback.
Refer: https://github.com/postmanlabs/postman-request/blob/master/tests/server.js#L47

test/fixtures/server.js
```diff
@@ -67,6 +70,49 @@ function createRawEchoServer () {
...
+        serverCaPath = path.join(certDataPath, 'ca-crt.pem'),
+
+        server = https.createServer({
+            key: fs.readFileSync(serverKeyPath),
+            cert: fs.readFileSync(serverCertPath),
+            ca: fs.readFileSync(serverCaPath),
+            requestCert: requestCert
+        });
+
+    server.on('request', function (req, res) {
```
codenirvana: Handling all the requests in a similar manner, this won't scale well.
The aim of this server(fixtures) is to remove the boilerplate code from tests but still be able to extend its usage.

vikiCoder: Then how about I take the request handler as an argument too. Like below....
```javascript
function createSSLServer(opts, reqHandler){
    if(!reqHandler){
        reqHandler = function defaultReqHandler(){
            ...
        }
    }

    server.on('request', reqHandler);

}
```

PR #723, opened 2019-01-30
Handle undefined in interrupted script execution

codenirvana:
Refer: https://github.com/postmanlabs/postman-app-support/issues/5799

### DIFF
--- a/lib/runner/extensions/event.command.js
+++ b/lib/runner/extensions/event.command.js
@@ -352,9 +352,11 @@ module.exports = {
                     result && result.response && (result.response = new sdk.Response(result.response));
 
                     // persist the pm.variables for the next script
-                    payload.context._variables = new sdk.VariableScope(result._variables);
+                    result && result._variables &&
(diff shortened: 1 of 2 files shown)


lib/runner/extensions/event.command.js
```diff
@@ -352,9 +352,11 @@ module.exports = {
                     result && result.response && (result.response = new sdk.Response(result.response));
 
                     // persist the pm.variables for the next script
-                    payload.context._variables = new sdk.VariableScope(result._variables);
+                    result && result._variables &&
+                        (payload.context._variables = new sdk.VariableScope(result._variables));
```
kunagpal: The type safety should be handled within the Collection SDK.

shamasis: Not future nitpick. I think that’s what should be done now. 

new VariableScope should just work without arguments. There was a test for this IMHO. 

Is the “result” object missing or is the bug in variable scope constructor?

PR #729, opened 2019-02-04
Added time option for requester to get detailed timings in response

vikiCoder:
Added time option to get detailed timing information about different phases of a request. Now you can use {time:<Boolean>} (true by default) in requester options for the same. The timing information will be added in response object returned in Requester.request function's callback.

**Note:** Need to merge https://github.com/postmanlabs/postman-collection/pull/796 first and bump SDK version(beta) for green builds. ✅

- [x] Bump `postman-collection` version(beta)

### DIFF
--- a/lib/requester/core.js
+++ b/lib/requester/core.js
@@ -208,6 +208,7 @@ module.exports = {
      * @param defaultOpts.followOriginalHttpMethod
      * @param defaultOpts.maxRedirects
      * @param defaultOpts.removeRefererHeaderOnRedirect
+     * @param defaultOpts.time
      * @param protocolProfileBehavior
      * @returns {{}}
      */
--- a/lib/requester/core.js
+++ b/lib/requester/core.js
@@ -208,7 +208,7 @@ module.exports = {
      * @param defaultOpts.followOriginalHttpMethod
      * @param defaultOpts.maxRedirects
      * @param defaultOpts.removeRefererHeaderOnRedirect
-     * @param defaultOpts.responseTimings
+     * @param defaultOpts.timings
      * @param protocolProfileBehavior
      * @returns {{}}
      */
@@ -234,7 +234,7 @@ module.exports = {
         options.jar = defaultOpts.cookieJar || true;
         options.timeout = defaultOpts.timeout;
         options.gzip = true;
-        options.time = defaultOpts.responseTimings;
+        options.time = defaultOpts.timings;
 
         // Ensures that "request" creates URL encoded formdata or querystring as
         // foo=bar&foo=baz instead of foo[0]=bar&foo[1]=baz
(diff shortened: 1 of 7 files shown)


lib/requester/core.js
```diff
@@ -208,6 +208,7 @@ module.exports = {
      * @param defaultOpts.followOriginalHttpMethod
      * @param defaultOpts.maxRedirects
      * @param defaultOpts.removeRefererHeaderOnRedirect
+     * @param defaultOpts.time
```
codenirvana: requires a more explicit name for our requester option which maps to postman-request option.
e.g removeRefererHeaderOnRedirect -> removeRefererHeader.

test/integration/requester-spec/timing.test.js
```diff
@@ -0,0 +1,77 @@
...
+                testrun = results;
+                done(err);
+            });
+        });
+
+        it('should include timing information by default', function () {
+            var response = testrun.response.getCall(0).args[2];
+
+            expect(response).to.have.property('timingStart');
+            expect(response).to.have.property('timings');
```
codenirvana: Check if this holds all the required properties.

lib/requester/requester.js
```diff
@@ -161,13 +162,19 @@ _.assign(Requester.prototype, /** @lends Requester.prototype */ {
                     acc.push(toChromeCookie(cookie));
                 }, []) : [];
 
+            if (res.timingStart) {
```
codenirvana:
```suggestion
            if (res && res.timingStart) {
```

lib/requester/requester.js
```diff
@@ -161,13 +162,19 @@ _.assign(Requester.prototype, /** @lends Requester.prototype */ {
...
+
             // Response in the SDK format
             response = new sdk.Response({ // @todo get rid of jsonifyResponse
                 code: responseJSON.statusCode,
                 status: res && res.statusMessage,
                 header: _.transform(responseJSON.headers, transformMultiValueHeaders, []),
                 stream: resBody,
-                responseTime: responseTime
+                responseTime: responseTime,
+                timingStart: res.timingStart,
```
codenirvana:
```suggestion
```

lib/requester/requester.js
vikiCoder: I think this is unnecessary. Because if res is undefined then there must be some error. In that case, it will return before reaching to this line.

codenirvana: Right, but as you can see the same is done on `L172` and since this information is coming from a different library, it's better to have this check.

test/integration/requester-spec/timing.test.js
```diff
@@ -0,0 +1,87 @@
...
+                testrun = results;
+                done(err);
+            });
+        });
+
+        it('should include timing information by default', function () {
+            var response = testrun.response.getCall(0).args[2];
+
+            expect(response).to.have.property('timings');
+            expect(response.timings).to.be.an('object').that.have.all.keys([
```
kunagpal: `.that.has`

PR #735, opened 2019-02-12
Allow request mutation in prerequest scripts

codenirvana:
Following properties can be mutated:
- url
- method
- headers


lib/runner/extensions/item.command.js
```diff
@@ -88,6 +97,13 @@ module.exports = {
                         });
                     }
 
+                    // update allowed request mutation properties with the mutated context
+                    // @note from this point forward, make sure this mutated
+                    // request instance is used for upcoming commands.
+                    ALLOWED_REQUEST_MUTATIONS.forEach(function (property) {
+                        item.request[property] = ctxTemplate.request[property];
```
shamasis: HeaderList has parent reference. You’d need to account for it. Check if there is any helper method to move lists.

test/integration/mutations/request.test.js
```diff
@@ -0,0 +1,355 @@
+var expect = require('chai').expect;
+
+describe('request mutations', function () {
```
kunagpal: All of these blocks need to have a test that checks for the pre-mutated state being valid.

johseffer:
When this change be released?

westonlast:
On both Postman 7.2.2 and PostmanCanary 7.3.4-canary01, I'm not able to add request parameters to the URL:

```
const Url = require("postman-collection").Url;  //where is this inclusion method even specified?

pm.request.url.addQueryParams({a: "a", b: "b"}); //does nothing

const url = new Url(pm.request.url.getRaw());
url.addQueryParams({
   a: "a",
   b: "b"
});
pm.request.url = url; //does nothing
```

PR #745, opened 2019-02-18
Move all change-logs to yankee

vikiCoder:


kunagpal:
@vikiCoder CHANGELOG.md also needs to be removed.

PR #754, opened 2019-02-21
Extends apikey authenticator to support multiple keys

ankitsiam:


lib/authorizer/apikey.js
```diff
@@ -28,6 +25,10 @@ module.exports = {
             version: '0.0.1'
         },
         updates: [
+            {
+                property: 'in',
```
shamasis: Not valid.

lib/authorizer/apikey.js
```diff
@@ -81,36 +83,42 @@ module.exports = {
      */
     sign: function (auth, request, done) {
         var target = TARGETS[auth.get('in')] || TARGETS.header,
-            key = auth.get('key'),
-            value = auth.get('value'),
-
-            lkey = _.lowerCase(key); // needed for header case insensitive matches
+            // in api key authentication method api keys are known by different names, thus we don't impose any
+            // default value for query name or header name and is provided by the user.There can be a single key or
+            // multiple keys.
```
shamasis: The members is not valid root level data structure

PR #755, opened 2019-02-21
Added `connection` system header

codenirvana:
- Add `connection` system header which is added by Node based on agent option.
- This is also useful for true request size calculation.
- Previously handled by https://github.com/postmanlabs/postman-collection/pull/814 hack.

### DIFF
--- a/lib/requester/requester.js
+++ b/lib/requester/requester.js
@@ -159,11 +159,11 @@ _.assign(Requester.prototype, /** @lends Requester.prototype */ {
                 return complete(err);
             }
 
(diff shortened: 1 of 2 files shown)


lib/requester/requester.js
```diff
@@ -229,6 +240,18 @@ _.assign(Requester.prototype, /** @lends Requester.prototype */ {
                 }
             });
 
+            connectionHeader = _.get(requestOptions, 'agentOptions.keepAlive');
```
kunagpal: `keepAlive`

PR #799, opened 2019-03-29
Support payload hash in hawk auth

vikiCoder:
- Added support for payload hash calculation in HAWK auth
Issue: https://github.com/postmanlabs/postman-app-support/issues/2458
- Fixed a bug in AWS auth where request execution will hangup if request has empty body


lib/authorizer/hawk.js
```diff
@@ -1,6 +1,10 @@
 var url = require('url'),
     _ = require('lodash'),
+    crypto = require('crypto'),
+    querystring = require('querystring'),
     Hawk = require('postman-request/lib/hawk'),
+    RequestBody = require('postman-collection').RequestBody,
+    urlencodedBodyBuilder = require('../requester/core-body-builder').urlencoded,
```
shamasis: bring from postman-encoder?

lib/authorizer/aws4.js
```diff
@@ -31,7 +31,7 @@ var _ = require('lodash'),
      * @param {Function} callback
      */
     computeBodyHash = function (body, algorithm, digestEncoding, callback) {
-        if (!(body && algorithm && digestEncoding)) { return callback(); }
+        if (!(body && algorithm && digestEncoding) || body.isEmpty()) { return callback(); }
```
shamasis: 👍

lib/authorizer/hawk.js
```diff
@@ -24,6 +28,84 @@ function randomString (length) {
...
+
+    if (body.mode === RequestBody.MODES.raw) {
+        hash.update(body.toString());
+        hash.update('\n');
+
+        return callback(hash.digest(digestEncoding));
+    }
+
+    if (body.mode === RequestBody.MODES.urlencoded) {
+        urlencodedBody = rfc3986(querystring.stringify(urlencodedBodyBuilder(body.urlencoded).form));
```
shamasis: too convoluted

lib/authorizer/hawk.js
vikiCoder: Will do that while making changes related to encoding in different PR.

lib/authorizer/hawk.js
vikiCoder: This will be removed once we use postman-url-encoder directly here. Will do that while making changes related to encoding in different PR.

PR #800, opened 2019-03-29
Handle GraphQL RequestBody

codenirvana:
- [x] Bump postman-collection
- [x] Add tests

### DIFF
--- a/lib/requester/core-body-builder.js
+++ b/lib/requester/core-body-builder.js
@@ -28,7 +28,11 @@ var _ = require('lodash'),
         json: 'application/json'
     },
 
-    FUNCTION = 'function',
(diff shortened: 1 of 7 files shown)


lib/requester/core-body-builder.js
```diff
@@ -148,5 +159,51 @@ module.exports = {
...
+                })
+            };
+        }
+
+        // otherwise, traverse the graphql properties and generate the
+        // stringified content. This avoids parsing the variables.
+        // @note this assumes that string content is escaped correctly
+        body = [];
+
+        GRAPHQL_PROPS.forEach(function (prop) {
```
shamasis: Why not JSON.stringify query and operation name and simply append variables?

PR #823, opened 2019-05-01
Fixed crashing of runtime because of invalid form-data value

vikiCoder:
- Added check on form-data value in body builder.

Solves:
https://github.com/postmanlabs/postman-app-support/issues/6180


lib/requester/core-body-builder.js
```diff
@@ -87,6 +88,12 @@ formDataBodyReducer = function (data, param) {
         },
         options; // we keep the default blank and then set to object wherever needed. saves doing object keyLength
 
+    // make sure that value is either string or read stream otherwise it'll cause error in postman-request
+    if (!(typeof formParam.value === 'string' ||
```
shamasis: I don’t think that detecting event emitter using instanceOf is a good idea. They’re horribly unreliable and there can be custom developed streams that has inherited event emitter in a non prototypal way. We should either detect for stream interface, like we did with read function here, or update postman-request to be more graceful with type mismatch.

vikiCoder: Do you mean detecting stream interface like below?
```javascript
if (!(typeof formParam.value === 'string' ||
      (
            typeof formParam.value.emit === 'function' &&
            typeof formParam.value.on === 'function' &&
            typeof formParam.value.read === 'function' &&
            typeof formParam.value.pause === 'function' &&
            typeof formParam.value.setEncoding === 'function' &&
            typeof formParam.value.resume === 'function' &&
            typeof formParam.value.pause === 'function'
      ))) {
        formParam.value = '';
}
```

vikiCoder:
Closing this in favour of https://github.com/postmanlabs/postman-runtime/pull/914

PR #832, opened 2019-05-15
Added followOrder lookup strategy to extract runnable items

codenirvana:
Added `followOrder` lookup strategy which supports:
- multiple id or name lookup
- enforce the order to follow
- supports duplicates i.e, iterate the same item multiple times

### DIFF
--- a/lib/runner/extract-runnable-items.js
+++ b/lib/runner/extract-runnable-items.js
@@ -3,6 +3,7 @@ var sdk = require('postman-collection'),
     Item = sdk.Item,
 
     DEFAULT_LOOKUP_STRATEGY = 'idOrName',
+    FOLLOW_ORDER_LOOKUP_STRATEGY = 'followOrder',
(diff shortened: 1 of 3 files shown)


lib/runner/extract-runnable-items.js
```diff
@@ -182,16 +195,63 @@ var sdk = require('postman-collection'),
...
+            entrypointLookup[entrypoints[i]] = true;
+        }
+
+        items = findItemsOrGroups(collection, entrypointLookup, true);
+
+        // at this point of time, we should have traversed all items mentioned in entrypoint and created a linear
+        // subset of items. However, if post that, we still have items remaining in lookup object, that implies that
+        // extra items were present in user input and corresponding items for those do not exist in collection. As such
+        // we need to bail out if any of the given entry-point is not found.
+        if (Object.keys(entrypointLookup).length) {
```
shamasis: I see what you mean now @codenirvana. even one extra item causes a bailout. but seems like that it is the acceptable product behaviour for now.

PR #839, opened 2019-05-31
Make payload hash in hawk auth optional

vikiCoder:
- Added `includePayloadHash` parameter to hawk auth

Issue: https://github.com/postmanlabs/postman-app-support/issues/2458#issuecomment-494320724


lib/authorizer/hawk.js
```diff
@@ -225,30 +226,34 @@ module.exports = {
...
-                nonce: params.nonce,
-                timestamp: params.timestamp,
-                ext: params.extraData,
-                app: params.app,
-                dlg: params.delegation,
-                user: params.user,
-                url: url.href, // force toString to add a protocol to the URL.
-                method: request.method,
-                hash: bodyHash
+        computeBodyHash(params.includePayloadHash && request.body, // pass body only when payload hash is required
```
shamasis: this `&&` approach does not differentiate between blank body versus option to not hash body.

codenirvana: @vikiCoder This will work but the intent is not clear here.
Instead, do an early bail out based on `includePayloadHash`.

test/fixtures/auth-requests.json
```diff
@@ -198,7 +199,8 @@
               "saveHelperData": true,
               "extraData": "skjdfklsjhdflkjhsdf",
               "app": "someAppId",
-              "delegation": ""
+              "delegation": "",
+              "includePayloadHash": true
```
codenirvana: If echo has no effect based on payload hash, it's better to have a local server which respect/throw-error based on the presence of hash.

lib/authorizer/hawk.js
```diff
@@ -210,45 +210,52 @@ module.exports = {
...
+                    timestamp: params.timestamp,
+                    ext: params.extraData,
+                    app: params.app,
+                    dlg: params.delegation,
+                    user: params.user,
+                    url: url.href, // force toString to add a protocol to the URL.
+                    method: request.method,
+                    hash: bodyHash
+                });
+                request.addHeader({
```
codenirvana: Add a newline above.

lib/authorizer/hawk.js
```diff
@@ -210,45 +210,52 @@ module.exports = {
...
+                });
+
+                return done();
+            };
 
         if (!params.authId || !params.authKey) {
             return done(); // Nothing to do if required parameters are not present.
         }
 
         request.removeHeader('Authorization', {ignoreCase: true});
```
codenirvana: Top level const for `Authorization`

lib/authorizer/hawk.js
```diff
@@ -210,45 +210,52 @@ module.exports = {
...
-                url: url.href, // force toString to add a protocol to the URL.
-                method: request.method,
-                hash: bodyHash
-            });
-            request.addHeader({
-                key: 'Authorization',
-                value: result,
-                system: true
-            });
+        if (!params.includePayloadHash) {
```
codenirvana: Add a brief description of why would someone opt-out or opt-in for payload hash.
And what's the consequence if we force include it for every time (existing behavior/bug)

test/fixtures/auth-requests.json
vikiCoder: Echo has support for payload verification now.

lib/authorizer/hawk.js
```diff
@@ -210,45 +211,61 @@ module.exports = {
...
             contentType = request.headers.get('content-type'),
 
             // force toString to add a protocol to the URL.
             url = urlEncoder.toNodeUrl(request.url.toString(true)),
 
-            self = this;
+            self = this,
+
+            signRequest = function (bodyHash) {
+                result = self.computeHeader({
```
codenirvana:
```suggestion
                var result = self.computeHeader({
```
Remove declaration from the parent function.

PR #869, opened 2019-07-24
Fixed freezing of collection run for GraphQL requests with AWS or Hawk auth

vikiCoder:
**What is the issue:**
The collection run completely freezes while sending GraphQL requests with AWS or Hawk authentication.

https://github.com/postmanlabs/postman-app-support/issues/6868

**Why is this happening:**

- In AWS and Hawk auth helpers, we calculate body hash by calling computeBodyHash function.
- Newly added body type GraphQL is not accounted in computeBodyHash function. So it never calls the callback for GraphQL body type.
- As a result entire collection run freezes!!

**Solution:**

- Add support for GraphQL body in computeBodyHash function of AWS and Hawk auth helpers.
- And call the callback if body type is unknown so that the same issue won’t happen again in future when we add some new body type.


lib/authorizer/aws4.js
```diff
@@ -78,6 +81,16 @@ var _ = require('lodash'),
...
+
+        if (body.mode === RequestBody.MODES.graphql) {
+            graphqlBody = bodyBuilder.graphql(body.graphql).body;
+            hash.update(graphqlBody);
+
+            return callback(hash.digest(digestEncoding));
+        }
+
+        // ensure that callback is called if body.mode doesn't match with any of the above modes
+        return callback();
```
codenirvana: with this, we can remove the `body.mode === RequestBody.MODES.formdata` check at the top.

PR #885, opened 2019-09-01
Fix abortOnError and stopOnError options for async scripts

vikiCoder:
**Isssue:**
- https://github.com/postmanlabs/postman-app-support/issues/6998

**Issue description:**
- `abortOnError` and `stopOnError` options of the Runtime should stop the collection/request execution if there are any errors while running scripts. But Runtime is not respecting these options when there is an error while running an asynchronous script.
- It is due to the fact that errors from the async script are not returned in the callback of `execute` function from Sandbox. Rather they are received in `execution.error` events.

**Solution:**
- Keep track of errors received from `execution.error` event and consider them while deciding to stop the run.


test/integration/runner-spec/abort-on-error-script.test.js
```diff
@@ -0,0 +1,56 @@
+var expect = require('chai').expect;
+
+describe('Abort on error', function () {
+    describe('Async script', function () {
+        var testrun;
+
+        before(function (done) {
+            this.run({
+                abortOnError: true,
```
codenirvana: Test with `abortOnError: false`

test/integration/runner-spec/stop-on-error-script.test.js
```diff
@@ -0,0 +1,55 @@
+var expect = require('chai').expect;
+
+describe('Stop on error', function () {
+    describe('Async script', function () {
+        var testrun;
+
+        before(function (done) {
+            this.run({
+                stopOnError: true,
```
codenirvana: Test with `stopOnError: false`

lib/runner/extensions/event.command.js
```diff
@@ -453,6 +466,12 @@ module.exports = {
                         }
                     });
 
+                    // Handle errors from async script. If there was an error running the script itself,
+                    // that takes precedence
+                    if (!err) {
```
codenirvana: Move this just after removing the event listeners.

The error needs to be serialized and all that is handled above.

lib/runner/extensions/event.command.js
```diff
@@ -288,6 +289,18 @@ module.exports = {
                         });
                     });
 
+                // To store error event, but only if needed. Because error in callback of host.execute()
+                // don't show execution errors for async scripts
+                (stopOnScriptError || abortOnError) &&
+                    this.host.on(EXECUTION_ERROR_EVENT_BASE + executionId, function (scriptCursor, err) {
```
codenirvana: `once` to store only the first async error.
Also, add a test for multiple async errors.

PR #918, opened 2019-10-23
Fix crashing of runtime for binary request body with AWS/Hawk/EdgeGrid auth

vikiCoder:
Ref: https://github.com/postmanlabs/postman-app-support/issues/7220


vikiCoder:
> can we have a test with hawk or edgegrid?

Added test with binary body in EdgeGrid auth.

PR #931, opened 2019-11-19
Add support for redirects in NTLM auth requests

coditva:
Tested with NTLM server setup locally.


coditva:
Support already present when `followAuthorizationHeader` option is set. Closing because PR is invalid.

PR #933, opened 2019-11-20
Update NTLM auth to find correct type2 header before parsing

coditva:
Also added a local NTLM server which sends `www-authenticate: Negotiate` along with type2 message. Adding this takes care of this bug.

### DIFF
--- a/lib/authorizer/ntlm.js
+++ b/lib/authorizer/ntlm.js
@@ -100,7 +100,8 @@ module.exports = {
             password = auth.get(NTLM_PARAMETERS.PASSWORD) || EMPTY,
             negotiateMessage, // type 1
             challengeMessage, // type 2
-            authenticateMessage; // type 3
+            authenticateMessage, // type 3
+            ntlmType2Header;
 
         if (response.code !== 401 && response.code !== 403) {
             return done(null, true);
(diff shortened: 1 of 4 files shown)


lib/authorizer/ntlm.js
```diff
@@ -130,7 +131,20 @@ module.exports = {
         }
         else if (state === STATES.T1_MSG_CREATED) {
             // At this point, we can assume that the type 1 message was sent to the server
-            challengeMessage = ntlmUtil.parseType2Message(response.headers.get(WWW_AUTHENTICATE) || EMPTY, _.noop);
+
+            // there can be multiple headers present with key `www-authenticate`.
+            // iterate to get the one which has the NTLM hash. if multiple
+            // headers have the NTLM hash, use the first one.
+            ntlmType2Header = response.headers.find(function (header) {
+                return header.key.toLowerCase() === WWW_AUTHENTICATE &&
```
codenirvana:
```suggestion
                return String(header.key).toLowerCase() === WWW_AUTHENTICATE &&
```

lib/authorizer/ntlm.js
```diff
@@ -130,7 +131,20 @@ module.exports = {
...
+                    header.valueOf().startsWith('NTLM ');
+            });
+
+            if (!ntlmType2Header) {
+                return done(new Error('ntlm: server did not send NTLM type2 message'));
```
codenirvana:
```suggestion
                return done(new Error('ntlm: server did not send NTLM type 2 message'));
```

PR #936, opened 2019-11-24
Pass object instead of string in 'proxy'  option of postman-request

vikiCoder:
Solves: postmanlabs/postman-app-support#6227

Todo:
- [ ] Update postman-request
- [ ] Update/Add Tests


vikiCoder:
Implemented different solution for this bug: https://github.com/postmanlabs/postman-collection/pull/958

PR #949, opened 2019-12-30
Avoid OOM in `normaliseIterationData`

coditva:
### Description

If iteration count is more than the length of `data`, the `normaliseIterationData` function fills the `data` array with the last element in `data` such that `data.length = iterationCount`.

When iteration count is a large number (e.g. 99999999), run creation takes a lot of time and sometimes even fails with "Out of Memory" error because the JS Heap is exhausted.

This change avoids the above situation by simply setting the `data[iterationCount - 1]` element as the value to be repeated. Later, when iteration data is to be fetched, if the element at `data[iteration]` exists, we use it; if not, we use the last element of the `data`.

---

### Performance increase
For `iterationCount` 9999999, time for `run` creation:
#### before
- 3s 417ms
- 3s 403ms
- 3s 505ms

#### after
- 0s 2.1ms
- 0s 1.9ms
- 0s 1.9ms

### DIFF
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -1,4 +1,6 @@
 master:
+  new features:
+    - GH-949 Improved performance for large number of iterations
   chores:
     - GH-943 Added `nyc` and `codecov` for code coverage checks
     - GH-951 Added sanity check for edgegrid auth params in sign function
--- a/lib/runner/extensions/waterfall.command.js
+++ b/lib/runner/extensions/waterfall.command.js
@@ -139,7 +139,9 @@ module.exports = {
                 this.queue('item', {
                     item: item,
                     coords: coords,
-                    data: this.state.data[coords.iteration],
+                    data: this.state.data[coords.iteration] === undefined ?
+                        this.state.data[this.state.data.length - 1] :
+                        this.state.data[coords.iteration],
                     environment: this.state.environment,
                     globals: this.state.globals,
                     collectionVariables: this.state.collectionVariables,
--- a/lib/runner/extensions/waterfall.command.js
+++ b/lib/runner/extensions/waterfall.command.js
@@ -3,7 +3,8 @@ var _ = require('lodash'),
     VariableScope = require('postman-collection').VariableScope,
 
     prepareLookupHash,
-    extractSNR;
+    extractSNR,
+    getIterationData;
 
 /**
  * Returns a hash of IDs and Names of items in an array
@@ -39,6 +40,21 @@ extractSNR = function (executions, previous) {
     });
 
     return snr;
+};
+
+/**
+ * Returns the data for the given iteration
+ *
+ * @function getIterationData
+ * @param {Array} data - The data array containing all iterations' data
+ * @param {Number} iteration - The iteration to get data for
+ * @return {Any} - The data for the iteration
+ */
+getIterationData = function (data, iteration) {
+    // check if data exists. if not, return the last member from the data
+    if (data && data.length) {
+        return iteration in data ? data[iteration] : data[data.length - 1];
+    }
 };
 
 /**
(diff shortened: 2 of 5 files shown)


test/integration/runner-spec/iterationCount.test.js
```diff
@@ -0,0 +1,112 @@
...
+                    {foo: 'bar'}
+                ],
+                collection: {
+                    item: [{
+                        event: [{
+                            listen: 'prerequest',
+                            script: {
+                                exec: `
+                                    pm.test('should contain data', function () {
+                                        pm.expect(pm.variables.get('foo')).to.equal('bar');
```
vikiCoder: to be more specific: `pm.iterationData.get('foo')`

lib/runner/extensions/waterfall.command.js
```diff
@@ -41,6 +42,28 @@ extractSNR = function (executions, previous) {
...
+/**
+ * Returns the data for the given iteration
+ *
+ * @function getIterationData
+ * @param {Array} data - The data array containing all iterations' data
+ * @param {Number} iteration - The iteration to get data for
+ * @return {Any} - The data for the iteration
+ */
+getIterationData = function (data, iteration) {
+    if (!Array.isArray(data) || data.length < 1) {
```
codenirvana: Don't think this check is required at this point.
`data` should be normalized before initializing any command.

test/integration/runner-spec/iterationCount.test.js
```diff
@@ -0,0 +1,112 @@
...
+            expect(testrun).to.nested.include({
+                'done.calledOnce': true,
+                'start.calledOnce': true
+            });
+
+            expect(testrun.request.callCount).to.equal(3);
+        });
+    });
+
+    describe('with data items lesser than iteration count', function () {
```
codenirvana: @vikiCoder ensure we test for:
- data.length > iterationCount
- data.length < iterationCount
- data without iterationCount

lib/runner/extensions/waterfall.command.js
coditva: For the case when run is created not by `runner.run`, but by `new Run` by the user?

test/integration/runner-spec/iterationCount.test.js
```diff
@@ -0,0 +1,230 @@
...
+            it('should complete the run successfully', function () {
+                expect(testrun).to.be.ok;
+                expect(testrun.done.getCall(0).args[0]).to.not.exist;
+                expect(testrun).to.nested.include({
+                    'done.calledOnce': true,
+                    'start.calledOnce': true
+                });
+            });
+
+            it('should use last data value for iterations with no corresponding data', function () {
```
vikiCoder: Only assert for the last iteration here

vikiCoder: Aalso add `it('should run collection specified number of times')`

test/integration/runner-spec/iterationCount.test.js
```diff
@@ -0,0 +1,230 @@
...
+
+            it('should complete the run successfully', function () {
+                expect(testrun).to.be.ok;
+                expect(testrun.done.getCall(0).args[0]).to.not.exist;
+                expect(testrun).to.nested.include({
+                    'done.calledOnce': true,
+                    'start.calledOnce': true
+                });
+            });
+
```
vikiCoder: Also add it('should run collection specified number of times')

test/integration/runner-spec/iterationCount.test.js
```diff
@@ -0,0 +1,230 @@
...
+            it('should set iterations equal to the length of data', function () {
```
vikiCoder: divide into:
- `it('should run iterations equal to the length of data')`
- `it('should use proper data in each iterations')`

test/integration/runner-spec/iterationCount.test.js
coditva: What we're also testing here is that if there is an `undefined` in data value, we don't use the last element of `data`.
See `assertions[1]` test.

CHANGELOG.yaml
```diff
@@ -1,4 +1,6 @@
 master:
+  improvement:
+    - GH-949 Added support to create a run with more than 99999999 iterations
```
vikiCoder: Fixed performance issue while running very large number of iterations

PR #950, opened 2020-01-06
Ensure non-empty filename param for form-data

coditva:
Needs updated `postman-request` (which should be updated with latest `form-data`)


lib/requester/core-body-builder.js
```diff
@@ -108,6 +108,12 @@ formDataBodyReducer = function (data, param) {
         }
     }
 
+    // make sure `filename` param is sent for every file without `value`
+    // so that `filename=""` is added to content-disposition header in form data
+    if (param.type === 'file' && !param.value && typeof param.fileName !== 'string') {
+        param.fileName = E;
```
vikiCoder: should also set param.value='' to make sure it doesn't cause any problems down the line

PR #956, opened 2020-01-20
Add test for large iteration count

coditva:
Add a test that check if run creation is successful for large iteration count. Not checking if iterations are run because it'll take a looong time.


coditva:
Force pushed to restart Travis build.

PR #992, opened 2020-03-19
Update request-send-flow.md

Suhasnama:
Fix : Image link URL

### DIFF
--- a/docs/request-send-flow.md
+++ b/docs/request-send-flow.md
@@ -1,4 +1,4 @@
 Postman Runtime handles all the steps required to send a request. This flowchart
 provides an overview of the steps:
 
-![Request Flow](https://github.com/postmanlabs/postman-runtime/blob/develop/docs/img/request-flow.png?raw=true)
+![Request Flow](./img/request-flow.png)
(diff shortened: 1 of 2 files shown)


Suhasnama:
Hi @codenirvana ,

- Will change current `URL` to `relative path`
- `staticFiles` location in `.jsdoc-config.json` is correctly configured i.e.
```
"staticFiles": {
        "include": ["./docs/img"]
      }
```
so no need to change that.

Suhasnama:
Requested changes are applied!!

docs/request-send-flow.md
```diff
@@ -1,4 +1,4 @@
 Postman Runtime handles all the steps required to send a request. This flowchart
 provides an overview of the steps:
 
-![Request Flow](request-flow.png)
+![Request Flow](./img/request-flow.png)
```
codenirvana:
```suggestion
![Request Flow](/img/request-flow.png)
```

Suhasnama: Hi @codenirvana , 
Tried your suggested path `/img/request-flow.png` , it fails because it it treated as absolute path (i.e., root / img / request-flow.png) . The fix to problem is to slightly modify the `staticFiles` path to `./docs` . Which now renders `request-flow.png` at both locations.

Suhasnama: I have tested it , with above mentioned changes and it works perfectly fine. Updating the `.jsdoc-config.json`

PR #1008, opened 2020-05-11
Added support for bubbling up history object irrespective of request error

ritwijsinha:
Bubble up `history` object in request and response callback on error


lib/requester/requester.js
```diff
@@ -411,15 +411,21 @@ _.assign(Requester.prototype, /** @lends Requester.prototype */ {
...
+                response;
+
             if (err) {
                 // bubble up http errors
                 // @todo - Should we send an empty sdk Response here?
-                return onEnd(err);
+                //
+                // Also sending `history` object to enable consumers to
+                // reliably figure out if session was created, hence enabling them to
+                // know whether it was a Request Error or Response Error.
```
codenirvana, line 425: These business logic details are not required to commented here.
This is just ensuring that the `history` object is sent even in case of error.

test/integration/request-flow/request-response-error.test.js
```diff
@@ -0,0 +1,93 @@
+var expect = require('chai').expect,
```
codenirvana, line 1: Test for history object in `requester-spec` test suite... asserting its structure and existence in case of error and success.

test/fixtures/server.js
```diff
@@ -647,6 +647,26 @@ function createDigestServer (options) {
     return app;
 }
 
+function createResponseErrorServer () {
```
codenirvana, line 650: This is not required.

lib/requester/requester.js
```diff
@@ -411,15 +411,21 @@ _.assign(Requester.prototype, /** @lends Requester.prototype */ {
...
+                return onEnd(err, null, history);
```
codenirvana, line 426: There's a difference between `null` and `undefined`.

test/fixtures/server.js
ritwijsinha: How are we going to test for response error without creating a server?

codenirvana: Request or response error doesn't matter here.
We just need to make sure the history object is preset on error.

PR #1011, opened 2020-05-12
Added new features in OAuth1 helper

vikiCoder:
Following improvements are done in OAuth1.0 helper,

- Added support for following signature methods:
  1. HMAC-SHA512
  2. RSA-SHA1
  3. RSA-SHA256
  4. RSA-SHA512
- Implemented Body Hash extension: https://tools.ietf.org/id/draft-eaton-oauth-bodyhash-00.html
- Added support for following additional parameters defined in [OAuth 1.0a spec](https://oauth.net/core/1.0a)
  1. oauth_callback
  2. oauth_verifier

Todo:
- [x] Update `node-oauth1` version with following PR merged: https://github.com/postmanlabs/node-oauth1/pull/12


lib/authorizer/oauth1.js
```diff
@@ -273,48 +417,19 @@ module.exports = {
...
-                urlencodedBody.add(param);
-            });
-        }
-        else {
-            request.addQueryParams(signatureParams);
+            return done();
         }
 
-        return done();
+        computeBodyHash(request.body, hashAlgo, 'base64', function (bodyHash) {
```
codenirvana, line 428: isn't this adding body hash irrespective of `includeBodyHash` option?

vikiCoder: It'll only be executed if `includeBodyHash` is true. Otherwise function will return from if block above.

PR #1014, opened 2020-05-13
Pass agent option through to postman-request

gone-skiing:
This first of two PRs aimed to address the need to run postman tests via socks proxy. It is a long-standing request documented here: postmanlabs/postman-app-support#1321.

The proposed first (minimum viable) step is to allow teams that use newman as a library to pass their own HTTP agent through to postman request. Postman request already supports agent option. This would enable use of libraries like socks-proxy-agent to run tests through socks proxy.

Please let me know if you need additional information.

### DIFF
--- a/lib/requester/core.js
+++ b/lib/requester/core.js
@@ -342,6 +342,7 @@ module.exports = {
      *
      * @param request
      * @param defaultOpts
+     * @param defaultOpts.agent
      * @param defaultOpts.network
      * @param defaultOpts.keepAlive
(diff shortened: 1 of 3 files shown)


gone-skiing:
@codenirvana here is a cleaner PR for the same change.

codenirvana:
@gone-skiing I am making a release today and wanted this change to be part of that... so extended this PR at #1015. Hope you don't mind 😅

gone-skiing:
@codenirvana 👍

PR #1039, opened 2020-06-01
Added support for SHA256 algorithm in Digest auth

vikiCoder:
- Added support for following algorithms in Digest auth
  1. SHA-256
  2. SHA-256-sess
  3. SHA-512-256
  4. SHA-512-256-sess
- Removed `crypto-js` form dependencies
- Added `js-sha512` dependency for SHA-512-256 algorithm

Issue: https://github.com/postmanlabs/postman-app-support/issues/5265

### DIFF
--- a/lib/authorizer/digest.js
+++ b/lib/authorizer/digest.js
@@ -1,5 +1,5 @@
 var _ = require('lodash'),
-    crypto = require('crypto-js'),
+    crypto = require('crypto'),
     urlEncoder = require('postman-url-encoder'),
 
     EMPTY = '',
@@ -11,6 +11,7 @@ var _ = require('lodash'),
     AUTH = 'auth',
     COLON = ':',
     QUOTE = '"',
+    SESS = '-sess',
     AUTH_INT = 'auth-int',
     AUTHORIZATION = 'Authorization',
     MD5_SESS = 'MD5-sess',
@@ -26,6 +27,14 @@ var _ = require('lodash'),
     OPAQUE_EQUALS_QUOTE = 'opaque="',
     QOP_EQUALS = 'qop=',
     NC_EQUALS = 'nc=',
+    ALGO = {
+        MD5: 'MD5',
+        MD5_SESS: 'MD5-sess',
+        SHA_256: 'SHA-256',
+        SHA_256_SESS: 'SHA-256-sess',
+        SHA_512_256: 'SHA-512-256',
+        SHA_512_256_SESS: 'SHA-512-256-sess'
+    },
     AUTH_PARAMETERS = [
         'algorithm',
         'username',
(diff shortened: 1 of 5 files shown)


lib/authorizer/digest.js
```diff
@@ -1,5 +1,5 @@
 var _ = require('lodash'),
-    crypto = require('crypto-js'),
```
shamasis, line 2: Do we know why it was -js earlier and what is the implication of changing this?

lib/authorizer/digest.js
```diff
@@ -243,31 +262,54 @@ module.exports = {
...
+                algorithm = algorithm || ALGO.MD5;
+                hashAlgo = 'md5';
+                break;
+            case ALGO.SHA_512_256:
+            case ALGO.SHA_512_256_SESS:
+            default:
+                // Current Electron version(7.2.3) in Postman app uses OpenSSL 1.1.0
+                // which don't support `SHA-512-256`.
+                // @todo: change this when Electron is upgraded to a version which supports `SHA-512-256`.
+                throw new Error(`Unsupported digest algorithm: ${algorithm}`);
```
shamasis, line 288: Do we try the alternate algorithm / node module that you discovered?

lib/authorizer/digest.js
vikiCoder: `crypto-js` was used earlier because Node's `Crypto` is not supported in browser(for chrome app). But now we are anyway using Node's crypto in other auths for algorithms like `RSA` which is not there in `crypto-js`.

And if we need to support it in browser, Browserify takes care of it using [crypto-browserify](https://github.com/crypto-browserify/crypto-browserify). So now `crypto-js` seems redundant dependency.

lib/authorizer/digest.js
vikiCoder: Will use that after doing sanity, security checks.

vikiCoder:
I've used `js-sha512` as a fallback for `SHA-512-256` algorithm because it is significantly slower than Node's `crypto`. See the benchmark below:

> js-sha x 56.98 ops/sec ±0.39% (73 runs sampled)
> crypto x 268 ops/sec ±7.32% (71 runs sampled)

lib/authorizer/digest.js
```diff
@@ -92,6 +102,23 @@ function _getDigestAuthHeader (headers) {
...
+ *
+ * @param {String} data string to calculate hash
+ * @param {String} algorithm hash algorithm
+ * @returns {String} hex encoded hash of given data
+ */
+function getHash (data, algorithm) {
+    // Current Electron version(7.2.3) in Postman app uses OpenSSL 1.1.0
+    // which don't support `SHA-512-256`. Use external `js-sha512` module
+    // to handle this case.
+    if (algorithm === 'sha512-256' && !_.includes(crypto.getHashes(), 'sha512-256')) {
```
codenirvana, line 116: Move `!_.includes(crypto.getHashes(), 'sha512-256')` to top-level const and load `jsSHA = require('js-sha512')` in memory only if required.

vikiCoder: Done

PR #1059, opened 2020-07-27
Added `statusMessageEncoding` option

appurva21:
Added `statusMessageEncoding` option to decode status message with specific encoding. 
All the status messages will now be decoded using `utf8` character encoding.

### DIFF
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -1,6 +1,6 @@
 master:
-  new features:
-    - GH-1059 Added `statusMessageEncoding` option to decode status message with specific encoding
+  fixed bugs:
+    - GH-1059 Fixed decoding of status messages with utf8 characters by using `statusMessageEncoding` option
 
 7.26.2:
   date: 2020-07-13
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -1,3 +1,7 @@
+master:
+  new features:
+    - GH-1059 Added `statusMessageEncoding` option to decode status message with specific encoding
+
 7.26.2:
   date: 2020-07-13
   chores:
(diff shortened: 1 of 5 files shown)


test/integration/sanity/response-status-message.test.js
```diff
@@ -0,0 +1,64 @@
...
+                    }, {
+                        request: {
+                            method: 'POST',
+                            url: rawEchoServer.url + '/raw-echo',
+                            body: {
+                                mode: 'raw',
+                                raw: 'HTTP/1.1 200 Работает нормально\r\n\r\n'
+                            }
+                        }
+                    }]
```
codenirvana, line 33: Add another test where it force encode a latin1 escaped character sequence.
Also, add a comment in the assertion explaining why it's force converted to utf8.

lib/requester/core.js
```diff
@@ -415,6 +415,11 @@ module.exports = {
         // set encoding to null so that the response is a stream
         options.encoding = null;
 
+        // Re-encode status message using `utf8` character encoding in postman-request.
+        // This is done to  correctly represent status messages with characters that lie outside
+        // the range of `latin` encoding (which is the default encoding in which status message is returned)
```
codenirvana, line 420:
```suggestion
        // the range of `latin1` encoding (which is the default encoding in which status message is returned)
```

CHANGELOG.yaml
```diff
@@ -1,3 +1,7 @@
+master:
+  new features:
```
codenirvana: This is not a feature and `statusMessageEncoding` option is not exposed.

test/integration/sanity/response-status-message.test.js
```diff
@@ -0,0 +1,64 @@
...
+        });
+    });
+
+    after(function () {
+        rawEchoServer.destroy();
+    });
+
+    it('should have completed the run', function () {
+        expect(testrun).to.be.ok;
+        expect(testrun.request.getCall(0)).to.nested.include({'args[0]': null, 'args[2].code': 200});
```
codenirvana, line 49: Check this in the individual requests `it` block.

PR #1094, opened 2020-10-05
Add integration test for large file upload

lucifercr07:
Added integration test for file upload of 50 MB.

### DIFF
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -1,3 +1,7 @@
+master:
+  chores:
+    - GH-1094 Added integration test for large files upload
+
 7.26.7:
   date: 2020-10-07
   chores:
--- a/test/fixtures/servers/http.js
+++ b/test/fixtures/servers/http.js
@@ -48,4 +48,27 @@ httpServer.on('/custom-reason', function (req, res) {
     res.end();
 });
 
+httpServer.on('/upload', function (req, res) {
+    if (req.method === 'POST') {
+        let body = [];
+
+        req.on('data', function (data) {
+            body.push(data);
+        });
+
+        req.on('end', function () {
+            res.writeHead(200, {'content-type': 'application/json'});
+            let response = {
+                'received-content-length': Buffer.concat(body).byteLength
+            };
+
+            res.end(JSON.stringify(response));
+        });
+    }
+    else {
+        res.writeHead(200, {'content-type': 'text/plain'});
+        res.end('Okay!');
+    }
+});
+
 module.exports = httpServer;
--- a/test/integration/file-uploads/request-body.test.js
+++ b/test/integration/file-uploads/request-body.test.js
@@ -1,7 +1,8 @@
 var fs = require('fs'),
     expect = require('chai').expect,
     sinon = require('sinon'),
-    IS_BROWSER = typeof window !== 'undefined';
+    IS_BROWSER = typeof window !== 'undefined',
+    {Readable} = require('stream');
 
 describe('file upload in request body', function () {
     var testrun;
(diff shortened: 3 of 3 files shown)


test/integration/file-uploads/request-body.test.js
```diff
@@ -645,4 +650,65 @@ describe('file upload in request body', function () {
                 .to.equal('Binary file load error: file resolver interface mismatch');
         });
     });
+
+    (IS_BROWSER ? describe.skip : describe)('should upload large files correctly', function () {
+        var testUploadFile = 'test/fixtures/upload-file-large.json';
```
coditva, line 655: Nitpick: We follow a convention of naming a constant variable in all caps.
```suggestion
        var TEST_UPLOAD_FILE = 'test/fixtures/upload-file-large.json';
```

test/integration/file-uploads/request-body.test.js
```diff
@@ -645,4 +650,65 @@ describe('file upload in request body', function () {
...
+
+        afterEach(function () {
+            sh.rm('-rf', testUploadFile);
+        });
+
+        // eslint-disable-next-line mocha/no-sibling-hooks
+        before(function (done) {
+            this.enableTimeouts(false);
+            if (IS_DARWIN) {
```
coditva, line 664: Nitpick: We can separate individual blocks of code by a newline. e.g. in this case, the `if` block can start after a newline and a newline after the last `else if`.

coditva, line 664: This can be abstracted out into a helper function at the top of this file so that other tests can use it too.

Also, we can use a `switch` on `process.platform` I think.

test/integration/file-uploads/request-body.test.js
```diff
@@ -645,4 +650,65 @@ describe('file upload in request body', function () {
                 .to.equal('Binary file load error: file resolver interface mismatch');
         });
     });
+
+    (IS_BROWSER ? describe.skip : describe)('should upload large files correctly', function () {
```
coditva, line 654: `describe` name should not be an assertion. It's mostly a noun or verb.

test/integration/file-uploads/request-body.test.js
```diff
@@ -645,4 +650,65 @@ describe('file upload in request body', function () {
...
+            }
+            this.run({
+                fileResolver: fs,
+                collection: {
+                    item: [{
+                        request: {
+                            url: 'https://postman-echo.com/post',
+                            method: 'POST',
+                            body: {
+                                mode: 'file',
```
coditva, line 682: Are we not testing file upload in **form-data** and **url-encoded**?

coditva (changes requested):

test/integration/file-uploads/request-body.test.js
```diff
@@ -651,26 +671,16 @@ describe('file upload in request body', function () {
         });
     });
 
-    (IS_BROWSER ? describe.skip : describe)('should upload large files correctly', function () {
-        var testUploadFile = 'test/fixtures/upload-file-large.json';
-
+    (IS_BROWSER ? describe.skip : describe)('large file upload in request body', function () {
         afterEach(function () {
```
coditva, line 675: We're deleting the file in `afterEach` whereas creating it in `before`. If another test is added in this `describe` block, it will fail because the file will not be available.

Let's either use `beforeEach` and `afterEach` or use `before` and `after`.

coditva (changes requested):

codenirvana (changes requested):
Let's see if we can achieve this by overriding the file resolver and sending a dynamic read stream.

CHANGELOG.yaml
```diff
@@ -1,3 +1,8 @@
+master:
+    date: 2020-10-05
```
codenirvana, line 2:
```suggestion
```

`date` here represents the release date.

test/fixtures/servers/http.js
```diff
@@ -48,4 +48,26 @@ httpServer.on('/custom-reason', function (req, res) {
     res.end();
 });
 
+httpServer.on('/file-upload', function (req, res) {
+    if (req.method === 'POST') {
+        var body = '';
+
+        req.on('data', function (data) {
+            body += data;
```
codenirvana, line 56: Don't convert Buffer to a string, instead concat buffers and calculate using `byteLength`.

test/fixtures/servers/http.js
```diff
@@ -48,4 +48,26 @@ httpServer.on('/custom-reason', function (req, res) {
...
+        });
+
+        req.on('end', function () {
+            try {
+                res.setHeader('content-length', String(body.length));
```
codenirvana, line 61: `content-length` response header signifies the length of the response body but here response is empty.

test/fixtures/servers/http.js
```diff
@@ -48,4 +48,26 @@ httpServer.on('/custom-reason', function (req, res) {
...
+                res.end();
+            }
+            catch (err) {
```
codenirvana, line 64: In which case this `err` will happen?

test/integration/file-uploads/request-body.test.js
```diff
@@ -645,4 +645,81 @@ describe('file upload in request body', function () {
...
+        // eslint-disable-next-line mocha/no-sibling-hooks
+        before(function (done) {
+            this.run({
+                fileResolver: {
+                    stat: function (src, cb) {
+                        cb(null, {isFile: function () { return true; }, mode: 33188});
+                    },
+                    createReadStream: function () {
+                        // creating buffer of size 52428800 bytes corresponds to 50 MB
+                        return Buffer.alloc(52428800);
```
codenirvana, line 659: Why this magic number? 
```suggestion
                        return Buffer.alloc(50 * 1024 * 1024);
```

test/integration/file-uploads/request-body.test.js
```diff
@@ -645,4 +645,81 @@ describe('file upload in request body', function () {
...
+
+    (IS_BROWSER ? describe.skip : describe)('large file upload in request body', function () {
+        // eslint-disable-next-line mocha/no-sibling-hooks
+        before(function (done) {
+            this.run({
+                fileResolver: {
+                    stat: function (src, cb) {
+                        cb(null, {isFile: function () { return true; }, mode: 33188});
+                    },
+                    createReadStream: function () {
```
codenirvana, line 657: Why it's returning a buffer instead of a[.](https://gist.github.com/wpscholar/270005d42b860b1c33cf5ab25b37928a) stream?

test/integration/file-uploads/request-body.test.js
```diff
@@ -645,4 +645,81 @@ describe('file upload in request body', function () {
                 .to.equal('Binary file load error: file resolver interface mismatch');
         });
     });
+
+    (IS_BROWSER ? describe.skip : describe)('large file upload in request body', function () {
+        // eslint-disable-next-line mocha/no-sibling-hooks
+        before(function (done) {
+            this.run({
+                fileResolver: {
```
codenirvana, line 653: add a comment explaining why this custom file resolver needed.

codenirvana (changes requested):

test/integration/file-uploads/request-body.test.js
lucifercr07: Fixed

CHANGELOG.yaml
```diff
@@ -1,3 +1,7 @@
+master:
+    chores:
```
codenirvana, line 2: wrong indentation.

test/integration/file-uploads/request-body.test.js
```diff
@@ -1,7 +1,8 @@
 var fs = require('fs'),
     expect = require('chai').expect,
     sinon = require('sinon'),
-    IS_BROWSER = typeof window !== 'undefined';
+    IS_BROWSER = typeof window !== 'undefined',
```
codenirvana, line 4: Why it's not working in the browser?

test/fixtures/servers/http.js
```diff
@@ -48,4 +48,19 @@ httpServer.on('/custom-reason', function (req, res) {
     res.end();
 });
 
+httpServer.on('/file-upload', function (req, res) {
+    if (req.method === 'POST') {
```
codenirvana, line 52: Why it only supports `POST`?

test/fixtures/servers/http.js
```diff
@@ -48,4 +48,19 @@ httpServer.on('/custom-reason', function (req, res) {
...
+        let body = [];
+
+        req.on('data', function (data) {
+            body.push(data);
```
codenirvana, line 56: Buffer has a `concat` method.

test/fixtures/servers/http.js
```diff
@@ -48,4 +48,19 @@ httpServer.on('/custom-reason', function (req, res) {
     res.end();
 });
 
+httpServer.on('/file-upload', function (req, res) {
```
codenirvana, line 51: Since it's not parsing any file, this can just be `/upload`.

test/fixtures/servers/http.js
```diff
@@ -48,4 +48,19 @@ httpServer.on('/custom-reason', function (req, res) {
...
+    if (req.method === 'POST') {
+        let body = [];
+
+        req.on('data', function (data) {
+            body.push(data);
+        });
+
+        req.on('end', function () {
+            res.writeHead(200, {'content-type': 'text/plain'});
+            res.end('received-content-length:' + Buffer.concat(body).byteLength);
```
codenirvana, line 61: What kind of an API response is this?
Either return the bytes received or a JSON response.

test/integration/file-uploads/request-body.test.js
codenirvana: Because you can't re-use the same stream. Return a new stream here instead of pushing data to a common stream.

codenirvana (changes requested):
Also, move the integration test to `/test/integration/benchmark`.

test/fixtures/servers/http.js
codenirvana: This unnecessarily casts the Buffer to String.

test/fixtures/servers/http.js
codenirvana: Have you benchmarked? Share the results.

test/fixtures/servers/http.js
codenirvana: What I meant was method doesn't matter. Don't have method-specific checks.

PR #1130, opened 2021-03-05
Added option to allow disabling TLS v1.3

mayank3012jain:
Changed the maxVersion when TLSv1.3 is to be disable, since secureOptions does not support it in node 12

### DIFF
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -1,6 +1,6 @@
 master:
   new features:
-    - GH-1130 Added support to disable TLSv1.3 in tlsDisabledProtocols protocol profile behavior.
+    - GH-1130 Added support to disable TLSv1.3 in tlsDisabledProtocols protocol profile behavior
 
 7.26.10:
   date: 2021-01-03
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -1,3 +1,7 @@
+master:
+  new features:
+    - GH-1130 Added support to disable TLSv1.3 in tlsDisabledProtocols protocol profile behavior.
+
 7.26.10:
   date: 2021-01-03
   chores:
--- a/lib/requester/core.js
+++ b/lib/requester/core.js
@@ -439,7 +439,14 @@ module.exports = {
         // the SSL and TLS protocol versions to disabled during negotiation
         if (Array.isArray(protocolProfileBehavior.tlsDisabledProtocols)) {
             protocolProfileBehavior.tlsDisabledProtocols.forEach(function (protocol) {
-                options.secureOptions |= constants[SSL_OP_NO + protocol];
+                // since secure options doesn't support TLSv1.3 before Node 14
+                // @todo remove the if condition when we drop support for Node 12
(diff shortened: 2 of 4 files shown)


CHANGELOG.yaml
```diff
@@ -1,3 +1,7 @@
+master:
+  new features:
+    - GH-1130 Added support to disable TLSv1.3
```
codenirvana, line 3: where?
> ... in `tlsDisabledProtocols` protocol profile behavior.

lib/requester/core.js
```diff
@@ -439,7 +439,14 @@ module.exports = {
         // the SSL and TLS protocol versions to disabled during negotiation
         if (Array.isArray(protocolProfileBehavior.tlsDisabledProtocols)) {
             protocolProfileBehavior.tlsDisabledProtocols.forEach(function (protocol) {
-                options.secureOptions |= constants[SSL_OP_NO + protocol];
+                // since secure options doesn't support TLSv1.3 before Node 14
+                // @todo remove the if condition when we drop support for Node 12
+                if (protocol === 'TLSv1_3' && !constants[SSL_OP_NO + protocol]) {
+                    options.maxVersion = 'TLSv1.2';
```
codenirvana, line 445: Since tests didn't cover this, have you ensure this works in Node v14?

lib/requester/core.js
```diff
@@ -439,7 +439,14 @@ module.exports = {
...
+                }
+                else {
+                    options.secureOptions |= constants[SSL_OP_NO + protocol];
```
codenirvana, line 448: Why is this not covered by tests 🤔

test/integration/protocol-profile-behavior/tlsProtocols.test.js
```diff
@@ -977,7 +978,7 @@ var fs = require('fs'),
                 });
             });
 
-            (TLSv1_3_SUPPORTED ? describe.skip : describe)('with just SSLv23 enabled', function () {
+            (TLSv1_3_SUPPORTED ? describe : describe)('with just SSLv23 enabled', function () {
```
codenirvana, line 981: Don't think this condition is needed anymore.

codenirvana (changes requested):

lib/requester/core.js
mayank3012jain: Yes, checked in Node v14. Works correctly without the `if` condition.

lib/requester/core.js
mayank3012jain: There are tests that cover this. But the codecov build is failing. Still trying to figure out why this is happening.

CHANGELOG.yaml
```diff
@@ -1,3 +1,7 @@
+master:
+  new features:
+    - GH-1130 Added support to disable TLSv1.3 in tlsDisabledProtocols protocol profile behavior.
```
codenirvana, line 3:
```suggestion
    - GH-1130 Added support to disable TLSv1.3 in tlsDisabledProtocols protocol profile behavior
```

codenirvana (changes requested):

codenirvana (dismissed):

codenirvana (approved):

PR #1137, opened 2021-03-16
Fix: The order of items corrected, when multiple folder arguments are passed.

adityaofficial10:
Fixes issue: https://github.com/postmanlabs/newman/issues/2085
**Background of the issue**
Postman Runtime runs the items in the same order, irrespective of the order in which they appear in the entrypoint.
Due to this, Newman also shows the same behaviour.
Integrating this will correct the behaviour there as well.

**What does this PR do?**
This PR fixes the order in which the items are executed when multiple Id lookup strategy is used.
To ensure the correct order, the runnable items are sorted by the order in which they appear in the entrypoint.

**What changes have been made?**
- An additional sort algorithm has been placed into the extractRunnableItems module to make sure the order of items is correct.


shamasis (commented):
@adityaofficial10 Have we hypothesised what the effect of merging this will have on Postman app? I would think that this code path having an option is safer. 

And also have someone from @postmanlabs/runtime internally run app integration tests based on these changes.

adityaofficial10:
> @adityaofficial10 Have we hypothesised what the effect of merging this will have on Postman app? I would think that this code path having an option is safer.
> 
> And also have someone from @postmanlabs/runtime internally run app integration tests based on these changes.

I ran all the tests and they run perfectly.
I agree with you and propose an option --order-folders for the same.
Would it be fine this way?

lib/runner/extract-runnable-items.js
```diff
@@ -186,6 +186,11 @@ var sdk = require('postman-collection'),
 
         items = findItemsOrGroups(collection, entrypointLookup, true);
 
+        // Sorting the items by the order in which they appear as folder arguments.
```
saswatds, line 189: This comment is misleading,  In the previous step, we queried to get both items and groups. So in this step, we are not just sorting by folders aka groups, but also requests aka items.

lib/runner/extract-runnable-items.js
```diff
@@ -186,6 +186,11 @@ var sdk = require('postman-collection'),
...
+        items && items.members && items.members.sort(function (a, b) {
```
saswatds, line 190: The mere existence of the `members` property does not guaranty the existence of the `sort` function. `members` must be an array.

lib/runner/extract-runnable-items.js
```diff
@@ -186,6 +186,11 @@ var sdk = require('postman-collection'),
...
+            return entrypoints.indexOf(a.name) - entrypoints.indexOf(b.name);
```
saswatds, line 191: Entry points will not always be just names. They can have id's in which case the indexOf will fail and just return `-1`.

saswatds (changes requested):

saswatds:
I am thinking in lines of `--follow-order` or `--preserve-order`.

adityaofficial10:
@shamasis 
I have made most of the changes.
I have chosen --preserve-order as the name.
Kindly tell me if we can add this as an attribute to entrypoint.
That would make more sense.

shamasis:
Ok, we have moved quite far without referring our GSOC mentor. Let's circle back with your mentor's take on these before (and how) we settle in on the rest.

adityaofficial10:
@shamasis 
Actually this is not related to GSoC I suppose.😅
I just thought that we can have an option --preserve-order in Newman and pass it to runtime as an attribute to entrypoint.
It would keep the code clean.

shamasis:
Still, your mentor would be the best person to merge these if approved.

adityaofficial10:
Ok thanks @shamasis !
I will try to contact him and get this merged.

lib/runner/extract-runnable-items.js
```diff
@@ -1,4 +1,5 @@
 var sdk = require('postman-collection'),
+    _ = require('lodash'),
```
saswatds, line 2: I do not see any reason why we should start using lodash here. 
Are we saying we can't use `Array.isArray` or `Array.prototype.includes` cannot be used here? 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array

lib/runner/extract-runnable-items.js
```diff
@@ -186,9 +188,14 @@ var sdk = require('postman-collection'),
...
         items = findItemsOrGroups(collection, entrypointLookup, true);
 
-        // Sorting the items by the order in which they appear as folder arguments.
-        items && items.members && items.members.sort(function (a, b) {
-            return entrypoints.indexOf(a.name) - entrypoints.indexOf(b.name);
+        // Sorting the items and folders by the order in which they appear as folder/request arguments,
+        // only if specified in the entrypoint.preserveOrder
+        preserveOrder && items && items.members && _.isArray(items.members) && items.members.sort(function (a, b) {
+            // Getting the keys of the folder/request as lookups can be on the basis of either id or name.
+            let key1 = _.includes(entrypoints, a.name) ? entrypoints.indexOf(a.name) : entrypoints.indexOf(a.id),
```
saswatds, line 195: Time complexity to high here. We are doing an includes and then indexOf. Need some justification for such a slow operation.

saswatds (changes requested):

lib/runner/extract-runnable-items.js
adityaofficial10: @saswatds
It was an ESLint suggestion.
So I used it.

lib/runner/extract-runnable-items.js
adityaofficial10: @saswatds
Ut was required for both id and name lookup.
Could you please guide me regarding the correct way?

adityaofficial10:
@saswatds 
I have made the required changes.
Kindly check if its fine now..

ShahOdin:
thanks for doing this. will this allow for running a folder multiple times? or will it still ignore me if I specify the folders: `--foo --bar --foo` and only run each once?

adityaofficial10:
@ShahOdin 
Actually the default newman behaviour has not been changed.
So, if this was possible before, it will still work.

ShahOdin:
> So, if this was possible before, it will still work.

The existing behavior was:

- if I had a collection with two requests: `Foo` and `Bar` in order
- if I asked `--folder Bar --folder Foo` it would run: `Foo` -> `Bar`
- if I asked `--folder Bar --folder Foo --folder Bar` it would run: `Foo` -> `Bar`
- if I asked `--folder Foo --folder Bar --folder Foo` it would run: `Foo` -> `Bar`

The intuitive behavior, I would argue is for Newman to return:

- Bar -> Foo
- Bar -> Foo -> Bar
- Foo -> Bar -> Foo

ie, create a new list altogether, and not just sort based on the order specified. does that make sense? 🤔

adityaofficial10:
@ShahOdin 
Sorry for this, but I have only corrected the order of the folders when --preserve-order option is set.
But the multiple folders thing is not there in Newman right now I guess. 
So I can't proceed on that as runtime is the base of all Postman Utilities and making this change will break a lot of things..

saswatds (approved):
@codenirvana  Looks ok to me, approving it. You can take it forward from here and suggest any changes to public API if required.

adityaofficial10:
@saswatds 
Thanks for your support.

lib/runner/extract-runnable-items.js
```diff
@@ -186,6 +187,19 @@ var sdk = require('postman-collection'),
 
         items = findItemsOrGroups(collection, entrypointLookup, true);
 
+        // Sorting the items and folders by the order in which they appear as folder/request arguments,
+        // only if specified in the entrypoint.preserveOrder
+        preserveOrder && items && items.members && Array.isArray(items.members) && items.members.sort(function (a, b) {
```
codenirvana, line 192: Instead of `sort`, why don't we just traverse the `entrypoints` array and get the reference `Item`s from `items.reference`?

test/integration/entrypoint/preserveOrder-entrypoint.test.js
```diff
@@ -0,0 +1,24 @@
+const {expect} = require('chai');
+
+
+describe('Entrypoint attribute preserveOrder', function () {
+    var testrun;
+
+    before(function (done) {
+        this.run({
+            collection: 'test/fixtures/Sample.postman_collection.json',
```
codenirvana, line 9: Fixtures are hard to maintain and you have to open 2 files to understand the test.
So, we avoiding that for all the new tests.

I feel we don't need an integration test for this, unit tests should cover all the changes.

test/unit/extract-runnable-items.test.js
```diff
@@ -487,6 +487,40 @@ describe('extractRunnableItems', function () {
                 }
             );
         });
+
+        describe('Preserve Order', function () {
```
codenirvana, line 491:
```suggestion
        describe('with preserveOrder', function () {
```

test/unit/extract-runnable-items.test.js
```diff
@@ -487,6 +487,40 @@ describe('extractRunnableItems', function () {
...
+            it('should work correctly when not specified', function (done) {
```
codenirvana, line 492: Test description doesn’t tell what it's asserting.

e.g, `should follow the collection order when unset`.

test/unit/extract-runnable-items.test.js
```diff
@@ -487,6 +487,40 @@ describe('extractRunnableItems', function () {
...
+                    function (err, runnableItems) {
+                        expect(err).to.be.null;
+                        expect(runnableItems).to.have.length(2);
+                        expect(_.map(runnableItems, 'name')).to.eql(['F2.R1', 'R1']);
+                        done();
+                    }
+                );
+            });
+
+            it('should work correctly when specified', function (done) {
```
codenirvana, line 508: same as above.

codenirvana (changes requested):

lib/runner/extract-runnable-items.js
adityaofficial10: @codenirvana 
Sure, I agree with this actually.

test/integration/entrypoint/preserveOrder-entrypoint.test.js
adityaofficial10: Ok if thats the case I'll remove it.

test/unit/extract-runnable-items.test.js
adityaofficial10: Sure, this is appreciated.

test/unit/extract-runnable-items.test.js
adityaofficial10: @codenirvana 
I'll do the needful.

test/unit/extract-runnable-items.test.js
adityaofficial10: Sure.

adityaofficial10:
@codenirvana 
I've made the required changes.
Kindly look if it's fine now.
Also @saswatds I just can't dismiss your review.
Github is just too much informative😂

lib/runner/extract-runnable-items.js
```diff
@@ -186,6 +188,15 @@ var sdk = require('postman-collection'),
 
         items = findItemsOrGroups(collection, entrypointLookup, true);
 
+        // Extracting the items and folders by the order in which they appear as folder/request arguments,
+        // only if specified in the entrypoint.preserveOrder
+        if (preserveOrder && items && items.reference && Array.isArray(entrypoints)) {
```
codenirvana, line 193:
```suggestion
        if (preserveOrder) {
```

- `Array.isArray(entrypoints)` - Already checked at `L179`
- `items && items.reference` - `findItemsOrGroups` is a internal function, guarantees to return `{ members, reference }`

lib/runner/extract-runnable-items.js
```diff
@@ -186,6 +188,15 @@ var sdk = require('postman-collection'),
...
+            entrypoints.forEach((item) => {
+                itemsInOrder && itemsInOrder.push(items.reference[item]);
+            });
+            items.members = itemsInOrder;
```
codenirvana, line 197:
```suggestion
            items.members = entrypoints.map(function (ref) {
                return items.reference[ref];
            });
```

This can be simplified by using the `map` instead of swapping arrays.

lib/runner/extract-runnable-items.js
```diff
@@ -263,6 +274,7 @@ var sdk = require('postman-collection'),
      * @param {String} [entrypoint.execute] id of item or group to execute (can be name when used with `idOrName`)
      * @param {Array<String>} [entrypoint.path] path leading to the item or group selected (only for `path` strategy)
      * @param {String} [entrypoint.lookupStrategy=idOrName] strategy to use for entrypoint lookup [idOrName, path]
+     * @param {Boolean} [entrypoint.preserveOrder] Value to preserve the order of folders/items or not
```
codenirvana, line 277: Not sure about this JSDoc description... can be improved.

test/unit/extract-runnable-items.test.js
```diff
@@ -487,6 +487,40 @@ describe('extractRunnableItems', function () {
                 }
             );
         });
+
+        describe('With preserve Order', function () {
```
codenirvana, line 491:
```suggestion
        describe('with preserveOrder', function () {
```

test/unit/extract-runnable-items.test.js
```diff
@@ -487,6 +487,40 @@ describe('extractRunnableItems', function () {
...
+                    function (err, runnableItems) {
+                        expect(err).to.be.null;
+                        expect(runnableItems).to.have.length(2);
+                        expect(_.map(runnableItems, 'name')).to.eql(['F2.R1', 'R1']);
+                        done();
+                    }
+                );
+            });
+
+            it('should follow the actual order when unset', function (done) {
```
codenirvana, line 508: Misleading/wrong title.

codenirvana (changes requested):

adityaofficial10:
@codenirvana 
I've made the required changes..
Kindly check if it looks fine.
And sorry for the delay..

codenirvana (approved):

adityaofficial10:
Thanks @codenirvana for your help and the merging..

PR #1143, opened 2021-03-30
Updated history object to include headers

mayank3012jain:
Add request and response headers to history object. 
Update the `addMissingRequestHeaders()` function to add headers from history.
Fixed a bug where in case of redirects, the request object in callbacks, had headers which were not present in the first request.

### DIFF
--- a/lib/requester/requester.js
+++ b/lib/requester/requester.js
@@ -229,15 +229,18 @@ _.assign(Requester.prototype, /** @lends Requester.prototype */ {
              */
             addMissingRequestHeaders = function (headers) {
                 _.forEach(headers, function (header) {
-                    // update headers which gets overwritten by the requester
-                    if (OVERWRITTEN_HEADERS[header.key.toLowerCase()]) {
-                        let lowerCasedKey = header.key.toLowerCase();
+                    var lowerCasedKey = header.key.toLowerCase();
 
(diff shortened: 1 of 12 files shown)


lib/requester/requester.js
```diff
@@ -228,48 +228,25 @@ _.assign(Requester.prototype, /** @lends Requester.prototype */ {
...
+                    if (OVERWRITTEN_HEADERS[header.key.toLowerCase()]) {
+                        let lowerCasedKey = header.key.toLowerCase();
+
+                        // in case of duplicate headers, remove all the duplicates
+                        // just keep the last item in the list.
+                        if (Array.isArray(_.get(request.headers, ['reference', lowerCasedKey]))) {
+                            request.headers.remove(function (h) {
+                                return _.lowerCase(h.key) === lowerCasedKey && h !== header;
+                            });
+                        }
```
codenirvana, line 242: Let's remove all this logic and just add/update the header

```suggestion
                _.forEach(headers, function (header) {
                    // update headers which gets overwritten by the requester
                    if (OVERWRITTEN_HEADERS[_.lowerCase(header && header.key)]) {
```

codenirvana (changes requested):
Can we also add a test for the missing `content-length` header?

codenirvana (approved):

PR #1158, opened 2021-06-07
Fix OAuth1.0a Incorrect Signature for URL Encoded Params

abhijeetborole:
TASK: Fix the Incorrect Signature Generation for URLs with Encoded Query Parameters
Implementation: Updated the message being sent to `node-oauth1` for signature generation. Added query parameters to action of this message by adding path instead of pathname in `getOAuth1BaseURL`. Removed queryParams from allParams which is passed to `node-oauth1` for signature generation.

Added a function to re-encode the query parameters to RFC-3986 and update these in the request being sent. This function uses formDecode and percentEncode from `node-oauth1`. `formDecode` breaks the query string down and decodes it with `decodePercent`, so that we can encode it again with `percentEncode` to the RFC-3986 standards and update the request. It correspondingly removes parameters from the request and adds the newly encoded parameters.

### DIFF
--- a/lib/authorizer/oauth1.js
+++ b/lib/authorizer/oauth1.js
@@ -247,6 +247,29 @@ module.exports = {
     },
 
     /**
+     * Updates the encoding for query parameters in request to RFC-3986.
+     * decodeFrom here decodes breaks the query string down and does
+     * percent decoding on parameters too.
+     *
+     * @param {Request} request - request to update query parameters
+     * @param {Object} url - Node.js like url object
+     */
+    updateQueryParamEncoding: function (request, url) {
+        queryParams = url.query && oAuth1.decodeForm(url.query);
+
+        queryParams && _.forEach(queryParams, function (param) {
+            encodedKey = param[0] && oAuth1.percentEncode(param[0]);
+            encodedValue = param[1] && oAuth1.percentEncode(param[1]);
+
+            if (encodedKey){
+                // remove parameter from request and add the newly encoded parameter
+                request.removeQueryParams(param[0]);
+                request.addQueryParams([{key: encodedKey, value: encodedValue}])
+            }
+        });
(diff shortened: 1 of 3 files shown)


lib/authorizer/oauth1.js
```diff
@@ -246,6 +246,29 @@ module.exports = {
...
 
+    /**
+     * Updates the encoding for query parameters in request to RFC-3986.
+     * decodeFrom here decodes breaks the query string down and does
+     * percent decoding on parameters too.
+     *
+     * @param {Request} request - request to update query parameters
+     * @param {Object} url - Node.js like url object
+     */
+    updateQueryParamEncoding: function (request, url) {
```
codenirvana, line 257: We don't want to export this utility function from this module.
This should be a private function like `getOAuth1BaseUrl`.

lib/authorizer/oauth1.js
```diff
@@ -246,6 +246,29 @@ module.exports = {
...
+
+        queryParams && _.forEach(queryParams, function (param) {
+            var encodedKey = param[0] && oAuth1.percentEncode(param[0]),
+                encodedValue = param[1] && oAuth1.percentEncode(param[1]);
+
+            if (encodedKey) {
+                // remove parameter from request and add the newly encoded parameter
+                request.removeQueryParams(param[0]);
+                request.addQueryParams([{key: encodedKey, value: encodedValue}]);
+            }
```
codenirvana, line 268: query param in the request could be different from the `url.query` which results in duplicates.

So, we can just remove all the query params (`request.url.query.clear()`) and add them back.

codenirvana (changes requested):

codenirvana (changes requested):

codenirvana (approved):

PR #1165, opened 2021-07-31
Remove bail out behavior if one of the entrypoints does not exist

ceoro9:
If one of the provided entry points is invalid(probably it does exist), then we **SHOULDN'T** bail out.

Why? Because this is completely surprising behavior for API.

If I have a list of valid entry points(at least I think so, in reality just one of them can be invalid. Mb typo or anything else, whatever). Then I try to run them, I'll get nothing in result. Exactly nothing. No the error, just success saying nothing was run.

What am I supposed to think of? I see my correct entry points and they're not run. This is so confusing.

The predictable behavior in that case would be throw an error, saying one of your entry points does not exist. Or just ignore the invalid entry points. The correct choice depends on the executed operation. If it's **Get**(means result should be returned), then error is more natural. If it's **Find**(like searching for something), then ignoring of non-existent entry points is totally fine.

Since in our case we got a look up(_lookupByMultipleIdOrName_), then ignorance is our choice.

Btw making some warning about the invalid entry points would be great, do we have a logger or anything like this?


codenirvana:
Agree that this is an unexpected behavior but I would rather like to bail out early with an error than executing an invalid (with missing items) run.

ceoro9:
@codenirvana , I'm totally fine to throw an error. Thank you for response 🙏🏼 I ll modify the code right in current PR.

lib/runner/extract-runnable-items.js
```diff
@@ -198,9 +198,9 @@ var sdk = require('postman-collection'),
         // at this point of time, we should have traversed all items mentioned in entrypoint and created a linear
         // subset of items. However, if post that, we still have items remaining in lookup object, that implies that
         // extra items were present in user input and corresponding items for those do not exist in collection. As such
-        // we need to bail out if any of the given entry-point is not found.
+        // we need to bail out with an error if any of the given entry-point is not found.
         if (Object.keys(entrypointLookup).length) {
-            return callback(null, []);
+            return callback(new Error('Some entrypoints are not found'));
```
ceoro9, line 203: could you help me to throw a semantically correct error, that will be properly handled by the dependent packages?

ceoro9: I've noticed there is some kind of special syntax for error message. Should we pick something similar?

Screenshot 2021-08-07 at 12 53 27

codenirvana:
@ceoro9 Actually, there's an `abortOnError` runner option (refer to [this test](https://github.com/postmanlabs/postman-runtime/blob/develop/test/unit/runner.test.js#L43)) which you can set to make sure the run terminates with an error.

ceoro9:
@codenirvana , cool. I ll use it. Thank you. Any other comments on code?

PR #1218, opened 2022-06-24
When iterating a data file, loop back to the beginning...

ThePieBandit:
rather than repeat the last element. For example, take the following data file:

```
[
    {"inputData":"foo"},
    {"inputData":"bar"}
]
```

If I were to run a collection with the following GET API: https://postman-echo.com/get?foo1={{inputData}} using the runner, several things happen:

1. When uploading the data file, it sets the iteration count to the length of the data array, in this case, 2. 
2. If the collection is run like that, it will run once with foo and once with bar.
3. After setting the data file, the iteration count can be changed. This can be useful for data seeding into an environment. For example, it can be set to 5.

the current behavior is to run once with foo and 4 times with bar. For data seeding, this is not ideal. this enhancement causes the iteration count, if greater than the number of rows in the data, to return to the first index of the data file and continue to invoke the APIs in the collection.

image

### DIFF
--- a/lib/runner/extensions/waterfall.command.js
+++ b/lib/runner/extensions/waterfall.command.js
@@ -51,13 +51,8 @@ extractSNR = function (executions, previous) {
  * @return {Any} - The data for the iteration
  */
 getIterationData = function (data, iteration) {
-    // if iteration has a corresponding data element use that
-    if (iteration < data.length) {
-        return data[iteration];
-    }
-
-    // otherwise use the last data element
-    return data[data.length - 1];
+    // if iteration has a corresponding data element use that, otherwise, start from the beginning again.
+    return data[iteration % data.length];
 };
 
 /**


ThePieBandit:
Note: I tried to write a test, but was unsuccessful. I wasn't sure how I could just unit test the getIterationData method without exposing it. I tried 
```
var expect = require('chai').expect,
    Waterfall = require('../../lib/runner/extensions/waterfall.command.js');

describe('runner extensions', function () {
    describe('Waterfall', function () {

        describe('getIterationData', function () {
            describe('when iteration count is less than the size of the data', function () {
                it('should return the index of the data', function () {
                    var data = [
                        { a: 'b' },
                        { c: 'd' },
                        { e: 'f' }
                    ];

                    expect(Waterfall.getIterationData(data, 1)).to.eql({ c: 'd' });
                });
            });
            describe('when iteration count is more than the size of the data', function () {
                it('should loop back around to the beginning of the data', function () {
                    var data = [
                        { a: 'b' },
                        { c: 'd' },
                        { e: 'f' }
                    ];

                    expect(Waterfall.getIterationData(data, 2)).to.eql({ e: 'f' });
                    expect(Waterfall.getIterationData(data, 3)).to.eql({ a: 'b' });
                    expect(Waterfall.getIterationData(data, 5)).to.eql({ e: 'f' });
                    expect(Waterfall.getIterationData(data, 7)).to.eql({ c: 'd' });
                });
            });
        });
    });
});
```

PR #1252, opened 2022-12-08
Add JWT token auth

sathishkumar-thangavel11:
JWT token Auth added


lib/authorizer/jwt.js
```diff
@@ -0,0 +1,273 @@
...
+ * request:{
+ *  auth:{
+ *      type:'jwt',
+ *      jwt:{
+ *          algorithm: <string> - ALGORITHMS_SUPPORTED,
+ *          header: {} - <valid JSON Object>  | <valid JSON string>
+ *          payload: {} - <valid JSON Object> | <valid JSON string>
+ *          secretOrPrivateKey: <string> - secret for HS algorithms | privateKey for RS, PS, ES algorithms
+ *                                         privateKey is generated using ssh-keygen or open-ssl
+ *          tokenAddTo: <string> - possible values - authorizationHeader | queryParam,
```
codenirvana, line 183:
```suggestion
 *          addTokenTo: <string> - possible values - header | queryParam,
```

lib/authorizer/jwt.js
```diff
@@ -0,0 +1,273 @@
...
+ *          tokenPrefix: <string> - prefix added before jwt token in Authorization header
```
codenirvana, line 184:
```suggestion
 *          headerPrefix: <string> - prefix added before jwt token in Authorization header
```

codenirvana (changes requested):

lib/authorizer/jwt.js
```diff
@@ -0,0 +1,221 @@
...
+                signKey = passphrase ? {
+                    key: privateKey,
+                    passphrase: passphrase
+                } : privateKey;
+            }
+
+            // generate the token and add to request
+            return jwt.sign(payload, signKey, options, function (err, jwtToken) {
+                if (err) {
+                    return done(err);
```
coditva, line 208: Would this fail the request and stop it from executing? Is that expected?

lib/authorizer/jwt.js
```diff
@@ -0,0 +1,221 @@
...
+    // jwt key constants
+    BEARER_AUTH_PREFIX = 'Bearer',
+    ALGORITHM = 'algorithm',
+    HEADER = 'header',
+    HEADER_ALGORITHM = 'alg',
+    PAYLOAD = 'payload',
+    SECRET = 'secret',
+    IS_SECRET_BASE_64_ENCODED = 'isSecretBase64Encoded',
+    PRIVATE_KEY = 'privateKey',
+    PASSPHRASE = 'passphrase',
```
coditva, line 13: What are the differences between `secret`, `privateKey` and `passphrase`? Should we document these?

lib/authorizer/jwt.js
```diff
@@ -0,0 +1,221 @@
...
+
+            // RS,PS,ES Algorithms use private key for token generation
+            else {
+                const privateKey = auth.get(PRIVATE_KEY),
+                    passphrase = auth.get(PASSPHRASE);
+
+                signKey = passphrase ? {
+                    key: privateKey,
+                    passphrase: passphrase
+                } : privateKey;
```
coditva, line 202: What happens if `privateKey` is not set? What error message (if any) would be sent?

lib/authorizer/jwt.js
```diff
@@ -0,0 +1,221 @@
...
+                    // treat root level alg as source of truth & remove alg from header to avoid conflict
+                    header: header && _.omit(header, [HEADER_ALGORITHM])
+                };
+
+            // HS Algorithms use secret for token generation
+            if (HS_ALGORITHMS[options.algorithm]) {
+                const secret = auth.get(SECRET),
+                    isBase64 = auth.get(IS_SECRET_BASE_64_ENCODED);
+
+                signKey = isBase64 ? Buffer.from(secret, BASE64).toString(ASCII) : secret;
```
coditva, line 191: What happens if `secret` is not a string?

test/integration/auth-methods/jwt.test.js
```diff
@@ -0,0 +1,1664 @@
...
+                    }
+                };
+
+                this.run(runOptions, function (err, results) {
+                    testrun = results;
+                    done(err);
+                });
+            });
+
+            it('should completed the run', function () {
```
coditva, line 222: Typo:
```suggestion
            it('should complete the run', function () {
```

test/integration/auth-methods/jwt.test.js
```diff
@@ -0,0 +1,1664 @@
...
+    algorithms = Object.entries(algorithmsSupported);
+
+describe('jwt auth', function () {
+    let testrun;
+
+    // with invalid algorithm - root level
+    algorithms.forEach(([key]) => {
+        const { alg, signKey } = algorithmsSupported[key];
+
+        describe(`with invalid algorithm for ${alg} algorithm`, function () {
```
coditva, line 194: Why do we need to do this for each of the algorithms?

coditva, line 194: The describe block string doesn't seem to be easy to read.
```suggestion
        describe(`with invalid algorithm key with "${alg}" algorithm set in header`, function () {
```

test/integration/auth-methods/jwt.test.js
```diff
@@ -0,0 +1,1664 @@
...
+                const runOptions = {
+                    collection: {
+                        item: {
+                            request: {
+                                url: 'https://postman-echo.com/headers',
+                                auth: {
+                                    type: 'jwt',
+                                    jwt: {
+                                        algorithm: alg,
+                                        header: null,
```
coditva, line 259: We should also test the case where `header` is an object with extra keys but invalid `alg` key.
```suggestion
                                        header: {
                                        	alg: 'some other alg',
                                        	myKey: 'myVal'
                                        },
```

test/integration/auth-methods/jwt.test.js
```diff
@@ -0,0 +1,1664 @@
...
+            it('should completed the run', function () {
+                expect(testrun).to.be.ok;
+                expect(testrun).to.nested.include({
+                    'done.calledOnce': true,
+                    'start.calledOnce': true,
+                    'request.calledOnce': true
+                });
+            });
+
+            it('should not add Authorization header', function () {
```
coditva, line 340: We should also assert what error message is logged (if any).

test/integration/auth-methods/jwt.test.js
```diff
@@ -0,0 +1,1664 @@
...
+
+                request.headers.members.forEach(function (header) {
+                    if (header.key === 'Authorization') {
+                        jwtToken = header.value.split('Bearer ')[1];
+                    }
+                    headers.push(header.key);
+                });
+
+                expect(request.headers.members).to.include.deep.members([
+                    new Header({ key: 'Authorization', value: `Bearer ${jwtToken}`, system: true })
```
coditva, line 505: What is this test doing? It's taking out the token from the request headers and test it to itself?

lib/authorizer/jwt.js
```diff
@@ -0,0 +1,221 @@
...
+    jwt = require('jsonwebtoken'),
+
+    // jwt key constants
+    BEARER_AUTH_PREFIX = 'Bearer',
+    ALGORITHM = 'algorithm',
+    HEADER = 'header',
+    HEADER_ALGORITHM = 'alg',
+    PAYLOAD = 'payload',
+    SECRET = 'secret',
+    IS_SECRET_BASE_64_ENCODED = 'isSecretBase64Encoded',
```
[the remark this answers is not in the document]
sathishkumar-thangavel11: Yes @coditva  its a boolean property

test/integration/auth-methods/jwt.test.js
sathishkumar-thangavel11: Yes @coditva  the jwtToken is generated and added to request, we are taking it out to verify the signature & payload to make sure it is correctly generated

lib/authorizer/jwt.js
sathishkumar-thangavel11: Yes @coditva documented in the schema comments block

test/integration/auth-methods/jwt.test.js
coditva: This `expect` seems weird because we took out the token from the headers, then compared the headers with the value we got. Something like:
```javascript
token = val;

expect(val).to.equal(token);
```

This `expect` block will never fail, no?

sathishkumar-thangavel11: Yes @coditva agree that expect block to again is unnecessary will remove it and keep the verifying part alone

lib/authorizer/jwt.js
coditva: We might need to add custom error message here.

sathishkumar-thangavel11: Yes we can add custom error for invalid secret or private key and let the library handle error for other edge cases

lib/authorizer/jwt.js
```diff
@@ -217,8 +219,18 @@ module.exports = {
                 payload = trimmedPayload && JSON.parse(trimmedPayload);
             }
 
+            payload = payload || {};
+
+            payload.iat = auth.get(AUTH_KEYS.IAT);
```
coditva, line 224: @codenirvana Should we fail/throw if `payload.iat` is also set? Currently we're just overriding.

imabp: @coditva we can do it like.
`payload.iat = payload.iat ?? auth.get(AUTH_KEYS.IAT)`

package.json
```diff
@@ -45,6 +45,7 @@
     "handlebars": "4.7.7",
     "httpntlm": "1.8.1",
     "js-sha512": "0.8.0",
+    "jsonwebtoken": "8.5.1",
```
codenirvana, line 48: check if we can upgrade to v9
https://github.com/auth0/node-jsonwebtoken/compare/v8.5.1...v9.0.0

lib/runner/request-helpers-presend.js
```diff
@@ -161,21 +161,24 @@ module.exports = [
...
+                        'warn',
+                        'runtime~' + authType + '.auth: could not sign the request: ' + (err.message || err));
+                }
+            },
             authSignHook = function () {
                 try {
                     authHandler.sign(authInterface, context.item.request, function (err) {
-                        // handle all types of errors in one place, see catch block
-                        if (err) { throw err; }
+                        logError(err);
```
codenirvana, line 174: `done()` called on error?

sathishkumar-thangavel11: Yes @codenirvana  after **logError** **done()** will be called . Earlier we are throwing error and the try catch block only works for synchronous code . JWT token generation works async and thus handled the done inside the callback instead of throwing

package.json
sathishkumar-thangavel11: Done upgraded to v9

codenirvana (approved):

PR #1260, opened 2023-01-17
Return with an error if the passed entry-point is invalid

shreyas-panduranga:
Postman Runtime bails out on detection of anything invalid in the entrypoint. It does so without returning any error. Due to this we are not able to detect an error condition and fail the run.

This ticket is to change that to return an error with specific codes to indicate the exact error that has occurred.

This change will affect all the consumers of postman-runtime: newman, postman-cli, Postman App Runner and Monitoring Agent
We are making this change to get the error back in monitoring agent, but we have to make sure we do not break anything in the other consumers as well. Change of behaviour is expected in certain scenarios (--folder or -i), but it should not break anything.

Newman change of behaviour:
Screenshot 2023-01-19 at 5 26 18 PM

Postman CLI change of behaviour:
Screenshot 2023-01-19 at 6 22 55 PM

Since newman and postman-cli use different strategies for running, the error is coming out to be different on both.


lib/runner/errors/constants.js
```diff
@@ -0,0 +1,34 @@
...
+        genericError: -1,
+        invalidLookupStrategy: 1,
+        invalidEntryPoints: 2,
+        invalidFolderOrRequest: 3
+    },
+    messageGenerators = {
+        // eslint-disable-next-line no-unused-vars
+        invalidFolderOrRequest: ({ folderOrRequestName }) => {
+            // eslint-disable-next-line no-template-curly-in-string
+            return INVALID_FOLDER_OR_REQUEST.replace('${folderOrRequestName}', folderOrRequestName);
```
pitss-pm, line 25: Should we shift message generation into respective error class?
That way we won't need template variables

lib/runner/errors/constants.js
```diff
@@ -0,0 +1,16 @@
+const errors = {
+        invalidLookupStrategy: 'invalidLookupStrategy',
+        invalidEntryPoints: 'invalidEntryPoints',
+        invalidFolderOrRequest: 'invalidFolderOrRequest'
+    },
+    errorCodes = {
+        genericError: -1,
+        invalidLookupStrategy: 1,
```
shreyas-panduranga, line 8: Numbered assuming only one of these can occur at a time.

shreyas-panduranga:
@codenirvana Gentle reminder for the review of this PR.

codenirvana (dismissed):

codenirvana (approved):

PR #1261, opened 2023-01-17
Add Support to Refresh OAuth2.0 Tokens

abhijeetborole:
Add Support for Refreshing OAuth2.0 Tokens


lib/authorizer/oauth2.js
```diff
@@ -12,7 +12,8 @@ var _ = require('lodash'),
         'addTokenTo',
         'tokenType',
         'headerPrefix'
-    ];
+    ],
+    REFRESH_TOKEN_TIMEOUT_MS = 30000;
```
abhijeetborole, line 16: Should this be a part of the run config?

lib/authorizer/oauth2.js
```diff
@@ -58,6 +59,34 @@ module.exports = {
      * @param {AuthHandlerInterface~authPreHookCallback} done -
      */
     pre: function (auth, done) {
+        const authId = auth.get('authId'),
```
abhijeetborole, line 62: Can't think of a better name than `authId` :sweat_smile:

lib/runner/request-helpers-presend.js
```diff
@@ -189,7 +189,7 @@ module.exports = [
             return done();
         }
 
-        authInterface = createAuthInterface(auth, context.protocolProfileBehavior);
+        authInterface = createAuthInterface(auth, context.protocolProfileBehavior, run.options.requester);
```
coditva, line 192: Maybe we can just pass the specific things from `options.requester` to keep the interface restrictive?

lib/authorizer/auth-interface.js
```diff
@@ -8,10 +8,11 @@ var _ = require('lodash'),
  * @constructs AuthInterface
  * @param {RequestAuth} auth -
  * @param {Object} protocolProfileBehavior - Protocol profile behaviors
+ * @param {Object} requester - Run options for this run
  * @return {AuthInterface}
  * @throws {Error}
  */
-createAuthInterface = function (auth, protocolProfileBehavior) {
+createAuthInterface = function (auth, protocolProfileBehavior, requester) {
```
coditva, line 15: `requesterOptions` maybe?
```suggestion
createAuthInterface = function (auth, protocolProfileBehavior, requesterOptions) {
```

lib/authorizer/oauth2.js
```diff
@@ -58,6 +59,34 @@ module.exports = {
...
+            refreshTokenHelper = auth.get('refreshTokenHelper');
+
+        // Go into the refresh token flow ONLY IF we have all necessary information present
+        if (authId && refreshTokenHelper && _.isFunction(refreshTokenHelper.refreshToken)) {
```
coditva, line 66: Why not just bail out if `authId` or `refreshTokenHelper` is not set?
```suggestion

        if (!(authId && refreshTokenHelper && _.isFunction(refreshTokenHelper.refreshToken))) {
        	return done(null, Boolean(auth.get('accessToken')));
        }
```

lib/authorizer/oauth2.js
```diff
@@ -58,6 +59,34 @@ module.exports = {
...
+            const refreshTokenPromise = refreshTokenHelper.refreshToken(authId);
```
coditva, line 67: Why do we have a method `refreshToken` inside the `refreshTokenHelper`? Shouldn't `refreshTokenHelper` itself be a method?

lib/authorizer/oauth2.js
```diff
@@ -58,6 +59,34 @@ module.exports = {
...
+
+            return Promise.race([
+                refreshTokenPromise,
+                new Promise((resolve) => {
+                    // We add a timeout to the refresh token promise to ensure that the request is not
+                    // hung up on the refresh token call. If the refresh token call takes more than
+                    // REFRESH_TOKEN_TIMEOUT_MS, we abandon and move forward with the execution.
+                    setTimeout(() => {
+                        return resolve();
```
coditva, line 76: Why resolve? Wouldn't this cause the token to be reset with `undefined`?

lib/authorizer/oauth2.js
```diff
@@ -58,6 +59,34 @@ module.exports = {
...
+                    }, REFRESH_TOKEN_TIMEOUT_MS);
+                })
+            ]).then((accessToken) => {
+                accessToken && auth.set('accessToken', accessToken, true);
+            }).catch(() => {
+                // do nothing, we throw an error from app when refresh fails
+                // ? TODO: should we throw an error here? app handles logging to console
+                // ? anyway
```
coditva, line 84: Think about how should someone using `postman-runtime` as a library would handle it. How would `newman` handle it?

coditva (changes requested):

lib/authorizer/oauth2.js
abhijeetborole: Nope, we only update the token if there's something present. Else we proceed as we would normally before.

lib/authorizer/oauth2.js
abhijeetborole: I've kept an interface as we may end up needing to expose more stuff here, especially with requests to support custom refresh flows with custom oauth spec coming up.

lib/authorizer/oauth2.js
coditva: Wouldn't it make more sense to reject here since this is an error (request timed out)? Also, how are we communicating this to the client? Should the timeout be implemented in the handler?

abhijeetborole: Was wondering if we should handle this in client or runtime, I think it's safer to have a timeout in runtime, if we add refresh support to monitors/newman. We would need them to implement a timeout as well, better to handle it here.

No communication to client right now. Wasn't sure if we should add an event for this case.

lib/authorizer/auth-interface.js
```diff
@@ -34,12 +35,23 @@ createAuthInterface = function (auth, protocolProfileBehavior) {
...
+                }
+
                 paramVariable = auth.parameters().one(keys);
 
                 return paramVariable && paramVariable.get();
             }
             if (_.isArray(keys)) {
                 return _.transform(keys, function (paramObject, key) {
+                    if (key === 'refreshTokenHelper') {
+                        paramObject[key] = requesterOptions.refreshTokenHelper;
```
coditva, line 52:
```suggestion
                        paramObject[key] = requesterOptions.refreshTokenHelper;

                        return paramObject;
```

lib/authorizer/oauth2.js
coditva, line 16:
```suggestion
    DEFAULT_REFRESH_TOKEN_TIMEOUT_MS = 30000;
```

lib/authorizer/oauth2.js
```diff
@@ -58,7 +59,46 @@ module.exports = {
...
-        done(null, Boolean(auth.get('accessToken')));
+        const authId = auth.get('authId'),
+            refreshTokenHelper = auth.get('refreshTokenHelper');
+
+        // Go into the refresh token flow ONLY IF we have all necessary information present
+        if (!(authId && refreshTokenHelper && _.isFunction(refreshTokenHelper.refreshToken))) {
+            return done(null, Boolean(auth.get('accessToken')));
+        }
+
+        let refreshTokenObject = refreshTokenHelper.refreshToken(authId),
```
coditva, line 70:
```suggestion
        var refreshTokenObject = refreshTokenHelper.refreshToken(authId),
```

coditva, line 70: Why not return the promise itself?

```suggestion
        let promise = refreshTokenHelper.refreshToken(authId),
```

lib/authorizer/oauth2.js
```diff
@@ -58,7 +59,46 @@ module.exports = {
...
+            refreshTokenTimeout = refreshTokenHelper.timeout,
+            refreshPromises = [];
+
+        if (!refreshTokenObject || !refreshTokenObject.promise) {
```
coditva, line 74: Let's consistently return a promise to avoid this block. In case of error we can return a rejected promise.

lib/authorizer/oauth2.js
```diff
@@ -58,7 +59,46 @@ module.exports = {
...
+            return done(null, Boolean(auth.get('accessToken')));
+        }
+
+        refreshPromises.push(refreshTokenObject.promise);
+
+        // Don't add a timeout if the timeout is set to 0
+        // Use the default timeout if no timeout is provided.
+        if (refreshTokenTimeout !== 0) {
```
coditva, line 82:
```suggestion
        if (!Number.isFinite(refreshTokenTimeout)) {
            refreshTokenTimeout = REFRESH_TOKEN_TIMEOUT_MS;
        }

        if (refreshTokenTimeout !== 0) {
```

test/integration/auth-methods/oauth2.test.js
```diff
@@ -0,0 +1,157 @@
...
+                                    }
+                                },
+                                url: 'https://postman-echo.com/get',
+                                method: 'GET'
+                            }
+                        }
+                    },
+                    requester: {
+                        refreshTokenHelper: {
+                            refreshToken () {
```
coditva, line 30: Let's also check that `refreshToken` was called.

test/integration/auth-methods/oauth2.test.js
```diff
@@ -0,0 +1,157 @@
...
+            });
+
+            it('should have sent the request with the refreshed access token', function () {
+                var request = testrun.request.getCall(0).args[3];
+
+                expect(request.headers.get('Authorization')).to.eql('Bearer some-refreshed-access-token');
+            });
+        });
+
+        describe('when token refresh fails', function () {
```
coditva, line 82: Also need tests when timeout occurs. Should also check that cancel method is called on timeout.

coditva (changes requested):

README.md
```diff
@@ -143,6 +143,17 @@ runner.run(collection, {
                 agentOptions: { keepAlive: true, timeout: 399 }
             },
             https: new https.Agent({ keepAlive: true })
+        },
+
+        // helper to refresh oauth2 tokens during execution
+        refreshTokenHelper: {
+            timeout: 30000, // timeout for the refresh token request in ms, defaults to 30000 ms
+            refreshToken: async function (authId) {
```
coditva, line 151: Since everywhere else we're using callbacks, should we keep things consistent by using callbacks? Or do we want to transition towards promises? @codenirvana

test/integration/auth-methods/oauth2.test.js
```diff
@@ -0,0 +1,260 @@
...
+    var testrun,
+        refreshTokenFunction,
+        cancelRefreshTokenFunction,
+        delayFor3000ms = () => {
+            return new Promise((resolve) => {
+                setTimeout(() => {
+                    resolve();
+                }, 3000);
+            });
+        };
```
coditva, line 18: Tip: this can be made generic by accepting a timeout:
```suggestion
        delay = (timeout) => {
            return new Promise((resolve) => {
                setTimeout(() => {
                    resolve();
                }, timeout);
            });
        };
```

lib/authorizer/oauth2.js
```diff
@@ -58,7 +59,62 @@ module.exports = {
...
+        }).catch((err) => {
+            // ? TODO: should we throw an error here? app handles logging to console
+            // ? anyway
+            // Only cancel the refresh if the refresh token call timed out. Runtime is not
+            // responsible for cancelling the refresh token call if it failed externally.
+            if (!refreshTimedOut) {
+                return;
+            }
+
+            _.isFunction(refreshTokenHelper.cancelRefresh) && refreshTokenHelper.cancelRefresh(err);
```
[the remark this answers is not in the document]
abhijeetborole: We support only one token refresh at a time.

codenirvana (approved):

PR #1274, opened 2023-03-08
Added json comment stripping for content type application/json

ashuvyaspm45:
Added support to strip the comment from json string before sending the request, if the content type is `application/json`


lib/requester/core-body-builder.js
```diff
@@ -22,6 +22,7 @@
  *                /___/___^//
  */
 var _ = require('lodash'),
```
sathishkumar-thangavel11, line 24: Can we change this imports to `const` instead of `var` ?

lib/requester/core-body-builder.js
```diff
@@ -187,21 +188,35 @@ module.exports = {
      * @returns {Object}
      */
     raw (content, request) {
-        var contentLanguage = _.get(request, 'body.options.raw.language', 'text');
+        var contentLanguage = _.get(request, 'body.options.raw.language', 'text'),
```
sathishkumar-thangavel11, line 191: Can we use `const` if `contentLanguage` will not be changed?

lib/requester/core-body-builder.js
```diff
@@ -187,21 +188,35 @@ module.exports = {
...
         // Add `Content-Type` header from body options if not set already
-        if (request && !oneNormalizedHeader(request.headers, CONTENT_TYPE_HEADER_KEY)) {
+        if (request && !contentTypeHeader) {
             request.headers.add({
                 key: CONTENT_TYPE_HEADER_KEY,
                 value: CONTENT_TYPE_LANGUAGE[contentLanguage] || CONTENT_TYPE_LANGUAGE.text,
                 system: true
             });
+
+            contentTypeHeader = oneNormalizedHeader(request.headers, CONTENT_TYPE_HEADER_KEY);
```
sathishkumar-thangavel11, line 206:
```
if (request && oneNormalizedHeader(request.headers, CONTENT_TYPE_HEADER_KEY)) {
            contentTypeHeader = oneNormalizedHeader(request.headers, CONTENT_TYPE_HEADER_KEY);
}
```

Should above code need to be below this line 206 so that there wont be a need reassign it again in line  206

lib/requester/core-body-builder.js
```diff
@@ -187,21 +188,35 @@ module.exports = {
...
         }
 
         if (typeof content !== STRING) {
             content = JSON.stringify(content);
         }
 
+
+        if (contentTypeHeader &&
```
sathishkumar-thangavel11, line 214: Will `contentTypeHeader` be `undefined` in any valid case ?

lib/requester/core-body-builder.js
```diff
@@ -187,21 +188,35 @@ module.exports = {
...
+            typeof contentTypeHeader.value === STRING &&
+            contentTypeHeader.value.startsWith('application/json')) {
```
sathishkumar-thangavel11, line 216: Is there any existing constant variable already available to use for `'application/json'` . If not we can assign it to a constant and reference here

package.json
```diff
@@ -58,6 +58,7 @@
     "postman-sandbox": "4.2.3",
     "postman-url-encoder": "3.0.5",
     "serialised-error": "1.1.3",
+    "strip-json-comments": "3.1.1",
     "uuid": "8.3.2"
```
sathishkumar-thangavel11, line 62: I hope we are using 3.1.1 for common js support . Have we considered the other parameters like bugs, improvements, any security fixes , package size between 3.1.1  and 5.0.0

test/integration/request-body/json-with-comment.test.js
```diff
@@ -0,0 +1,169 @@
...
+    describe(test.suitName, function () {
+        let testrun, URL_HEADER;
+
+        before(function () {
+            URL_HEADER = global.servers.http + '/echo/post';
+        });
+
+        test.scenarios.forEach((scenario) => {
+            describe(test.scenario, function () {
+                before(function (done) {
```
sathishkumar-thangavel11, line 119: ` describe(test.scenario, function () {` will this print an object for the describe string ?

lib/requester/core-body-builder.js
ashuvyaspm45: One case is when request is not present. As there were already a check on request presence.

test/integration/request-body/json-with-comment.test.js
ashuvyaspm45: it's `test.scenario` not scenario. So I thought to have a default scenario name for test suit. Later can add specific to any scenario like `describe(scenario.name || test.scenario`

`
      {
            suitName: 'when single line comment present',
            scenario: 'it should remove comments',
            scenarios: [
                {
                    jsoncString: '{// some comment\n"a": "value"\n}',
                    expectedObject: { a: 'value' }
                },
`

lib/requester/core-body-builder.js
```diff
@@ -40,9 +41,10 @@ var _ = require('lodash'),
     },
 
     STRING = 'string',
-    E = '',
+    E = '';
 
-    oneNormalizedHeader,
+
+var oneNormalizedHeader,
```
sathishkumar-thangavel11, line 47: can we use let or const ?

lib/requester/core-body-builder.js
```diff
@@ -198,10 +201,21 @@ module.exports = {
...
 
+        if (request && oneNormalizedHeader(request.headers, CONTENT_TYPE_HEADER_KEY)) {
+            contentTypeHeader = oneNormalizedHeader(request.headers, CONTENT_TYPE_HEADER_KEY);
+        }
+
         if (typeof content !== STRING) {
             content = JSON.stringify(content);
         }
 
+
```
sathishkumar-thangavel11, line 212: ```const contentTypeHeader  =  oneNormalizedHeader(request.headers, CONTENT_TYPE_HEADER_KEY)```

Can we assign this here and remove line about if block ?

lib/runner/request-helpers-presend.js
```diff
@@ -381,5 +383,31 @@ module.exports = [
...
             done();
         });
+    },
+    // Raw request body update
+    function (context, run, done) {
+        if (!context.item) { return done(new Error('Nothing to update body.')); }
+
+        let request = _.get(context.item, 'request'),
+            requestBody = _.get(request, 'body'),
+            bodyParams;
```
sathishkumar-thangavel11, line 393: Can we use ```lib/runner/request-helpers-presend.js``` as source for removing JSON comments and leave core body builder untouched ? @codenirvana  any thoughts on this ?

ashuvyaspm45:
The current implementation does the request body formatting (removing comments) in `core-body-builder.js`, as it is the module build for request body transformation.

Now with the requirement to mutate the request body, so it can have the raw data with comments removed, we are doing that in `request-helpers-present.js` to have this processing right before we are sending the request and triggers after that will have the `raw` data with comments removed. 

In current solution the comment sanitisation logic is called twice and when it is called the second time the body has no comments present as it is already being removed in `present`.

Based on the above analysis we can move the comment removal step in `present` and remove from `core-body-builder`. 

Also the `core-body-builder.js` body formatting method is also being called from auth handlers to compute the body hash. 
Which get triggers from: `present` and `dryrun`
With moving the logic to present we will cover the case of auth handler as well. Still not sure about the use case of `dryrun`. 
@codenirvana Any thoughts on this?

sathishkumar-thangavel11 (commented):
LGTM , Lets test across all agents

package.json
ashuvyaspm45: Checked there were no bug fixes, just additional functionality of trimming comas and making this esm module.

lib/runner/request-helpers-presend.js
```diff
@@ -1,15 +1,63 @@
...
+
+        // if body is not available
+        if (!requestBody || requestBody.isEmpty()) {
+            return done();
+        }
+
+        // if data mode is not raw bail out
+        if (requestBody.mode !== RequestBody.MODES.raw) {
+            return done();
+        }
```
codenirvana, line 34: Not required.

test/fixtures/servers/http.js
```diff
@@ -61,4 +61,28 @@ httpServer.on('/redirect-to', function (req, res) {
     res.end();
 });
 
+httpServer.on('/echo/post', function (req, res) {
```
codenirvana, line 64:
```suggestion
httpServer.on('/echo', function (req, res) {
```

test/fixtures/servers/http.js
```diff
@@ -61,4 +61,28 @@ httpServer.on('/redirect-to', function (req, res) {
...
+            body = JSON.parse(body);
+        }
+        // We want to return the parsed body if the incoming request body
+        // is valid json string. If not we want to return it as incoming string.
+        // eslint-disable-next-line no-empty
+        catch (e) {}
+
+        res.writeHead(200, { 'Content-Type': 'application/json' });
+
+        const responseBody = { data: body, headers: headers, url: url };
```
codenirvana, line 83: Send raw body in `data` and parsed JSON in `json` prop.

PR #1312, opened 2023-06-13
Fixed dns resolution for localhost failing when node version >= v20

KiraElric007:
Ref: https://github.com/postmanlabs/newman/issues/3104

Ran the test suite with node version 20 locally. I verified that this issue is fixed.
But also found out that this not the only issue with upgrading to node 20:
-> TLSv1 and TLSv1_1 related tests are failing. (other versions of TLS are passing)
-> certificate related tests are failing.


test/unit/requester-core.test.js
```diff
@@ -270,6 +284,40 @@ describe('requester util', function () {
             expect(requesterCore.getRequestOptions(request, options).lookup).to.be.a('function');
         });
 
+        it('should override lookup function and resolve localhost correctly when dnsOpts.all is true', function () {
+            var request = new sdk.Request({
+                    url: 'http://localhost:8080/random/path'
```
nikithaachhini23, line 289: We shall even add a test case when dnsOpts.all is false - then it picks up a single address string

lib/requester/core.js
```diff
@@ -289,6 +291,20 @@ var dns = require('dns'),
...
+            // When options.all is set, the callback expects array of addresses instead of a single address
+            if (options.all) {
+                return _lookup(options, hostLookup, lowercaseHost, function (err, addresses) {
+                    if (err) { return callback(err); }
+
+                    addresses = _.filter(addresses, (addr) => {
+                        return !self.isAddressRestricted(addr, networkOpts);
+                    });
+
+                    return callback(_.isEmpty(addresses) ?
```
nikithaachhini23, line 303: For error why are we checking if addresses exist or not?

lib/requester/core.js
```diff
@@ -229,7 +229,9 @@ var dns = require('dns'),
             // returning error synchronously causes uncaught error because listeners are not attached to error events
             // on socket yet
             return setImmediate(function () {
-                callback(null, resolvedAddr, resolvedFamily);
+                // When options.all is set, the callback expects array of addresses instead of a single address
+                return options.all ? callback(null, [{ address: resolvedAddr, family: resolvedFamily }]) :
```
nikithaachhini23, line 233: We're just sending the available address and family as an array of object. Don't we have list of addresses?

KiraElric007: network.hostLookup option does not support multiple IPs at the moment. That is a separate feature improvement we can do after this PR maybe.

lib/requester/core.js
KiraElric007: For a host name if no ip is available (empty list of addresses) we send `Error(ERROR_ADDRESS_RESOLVE + hostname)` ( `NETERR: getaddrinfo ENOTFOUND`)

test/unit/requester-core.test.js
KiraElric007: Yes. Already added it as part of already existing `it` statement. Line No: 219 and 240

lib/requester/core.js
aman-postman, line 233: where is this documented ? reference for options.all will help

lib/requester/core.js
```diff
@@ -289,6 +291,20 @@ var dns = require('dns'),
         // - localhost
         // - *.localhost
         if (getTLD(lowercaseHost) !== LOCALHOST) {
+            // When options.all is set, the callback expects array of addresses instead of a single address
+            if (options.all) {
```
aman-postman, line 295: was this breaking till now ?

test/unit/requester-core.test.js
```diff
@@ -3,6 +3,8 @@ var expect = require('chai').expect,
     runtimeVersion = require('../../package').version,
     requesterCore = require('../../lib/requester/core');
 
+const IS_NODE = typeof window === 'undefined';
```
aman-postman, line 6: in tests, when will this be false ?

test/unit/requester-core.test.js
```diff
@@ -270,6 +290,42 @@ describe('requester util', function () {
...
+            options.lookup(hostname, { all: true }, function (err, addresses) {
+                expect(err).to.be.null;
+                expect(addresses).to.be.an('array');
+                expect(addresses[0].address).to.be.oneOf(['127.0.0.1', '::1']);
+                expect(addresses[0].family).to.be.oneOf([4, 6]);
+            });
+        });
+
+        // eslint-disable-next-line max-len
+        (IS_NODE ? it : it.skip)('should override lookup function and resolve *.localhost correctly when dnsOpts.all is true', function () {
```
aman-postman, line 312: what is expectation from code when is node is false ? why not assert that instead of skipping

lib/requester/core.js
KiraElric007: No. Only after node version 20. Reason in RCA - https://postmanlabs.atlassian.net/wiki/spaces/AT/pages/4532502534/APITEST-375+Localhost+not+resolving+in+postman-runtime+for+node+version+20

lib/requester/core.js
KiraElric007: Ref: for options.all - https://www.geeksforgeeks.org/node-js-dns-lookup-method/
Also mentioned the same in RCA.

test/unit/requester-core.test.js
KiraElric007: Browser cannot connect to sockets. There are few tests that are not applicable for browser. This is how tests are return in postman-runtime in general. Also the generic expectation of what happens in browser is already covered in other test cases.

test/unit/requester-core.test.js
KiraElric007: For browser tests, this will be false. All the other tests this will be true.

aman-postman (approved):

bdoherty-tovala:
Hoping this fix can be merged soon as it is currently a blocker for me.

codenirvana (approved):

PR #1326, opened 2023-08-02
Added support for uploading binary files via base64 encoded string

codenirvana:
Closes #1327


coditva (approved):

lib/requester/core-body-builder.js
```diff
@@ -154,18 +154,13 @@ formDataBodyReducer = function (data, param) {
...
         (options || (options = {})).contentType = param.contentType;
     }
 
-    // additionally parse the file name and length if sent
-    // @note: Add support for fileName & fileLength option in Schema & SDK.
-    //        The filepath property overrides filename and may contain a relative path.
-    if (typeof param.fileName === STRING) { (options || (options = {})).filename = param.fileName; }
-    if (typeof param.fileLength === 'number') { (options || (options = {})).knownLength = param.fileLength; }
-
+    if (param.fileName && typeof param.fileName === STRING) {
```
dev-sharma-08, line 161: What will happen when fileName is not present ?

codenirvana: Not an issue, set as undefined.

PR #1358, opened 2023-11-30
feat: add new secret variable type

Pranav2612000:


lib/runner/extensions/event.command.js
```diff
@@ -11,8 +11,9 @@ var _ = require('lodash'),
     createItemContext = require('../create-item-context'),
 
     ASSERTION_FAILURE = 'AssertionFailure',
-    SAFE_CONTEXT_VARIABLES = ['_variables', 'environment', 'globals', 'collectionVariables', 'cookies', 'data',
-        'request', 'response'],
+    SAFE_CONTEXT_VARIABLES =
+        ['_variables', 'environment', 'globals', 'secrets', 'collectionVariables', 'cookies', 'data',
```
theonly1me, line 15: @Pranav2612000 let's call them `vaultVariables` similar to `collectionVariables`. Calling them `secrets` adds ambiguity because of the `secret` type that exists.

lib/runner/extensions/request.command.js
```diff
@@ -15,6 +15,7 @@ var _ = require('lodash'),
      * @param {VariableScope} payload.environment -
      * @param {VariableScope} payload.collectionVariables -
      * @param {VariableScope} payload.globals -
+     * @param {VariableScope} payload.secrets -
```
theonly1me, line 18: @Pranav2612000 here too let's keep it as `vaultVariables`

theonly1me (approved):

PR #1361, opened 2023-12-05
Fix - Auth state management for ntlm auth in runner

sandeep-singh-pstmn:
In this PR, we are storing the Auth state in a separate variable, which resets every time a new req item runs
after this chnage, we will not be able to access state data in the script, which makes sense as to why we need to expose internal state management in the script to store that in auth object

Screenshot 2023-12-05 at 10 19 30 AM


lib/authorizer/ntlm.js
```diff
@@ -150,7 +149,7 @@ module.exports = {
      * @param {AuthHandlerInterface~authPreHookCallback} done -
      */
     pre: function (auth, done) {
-        !auth.get(STATE) && auth.set(STATE, STATES.INITIALIZED);
+        !auth.getAuthState(NTLM) && auth.setAuthState(NTLM, STATES.INITIALIZED);
```
coditva, line 152: All auths should be using the state now instead of `auth.set`. Can we ensure that there's no use of `auth.set` for updating the state?

lib/authorizer/auth-interface.js
```diff
@@ -109,6 +110,30 @@ createAuthInterface = function (auth, protocolProfileBehavior, options) {
             });
 
             return this;
+        },
+
+        /**
+         * @param {String} type - type of auth
+         * @return {AuthState|undefined}
+         */
+        getAuthState: function (type) {
```
coditva, line 119: Since this is auth-interface, using `auth` in the name is redundant. It can simply be `getState` and `setState`.

lib/authorizer/auth-interface.js
```diff
@@ -109,6 +110,30 @@ createAuthInterface = function (auth, protocolProfileBehavior, options) {
             });
 
             return this;
+        },
+
+        /**
+         * @param {String} type - type of auth
+         * @return {AuthState|undefined}
```
coditva, line 117: Where is the `AuthState` type defined?

lib/authorizer/index.js
```diff
@@ -110,7 +110,8 @@ authorizeRequest = function (request, done) {
 
     var clonedReq = new sdk.Request(request.toJSON()),
         auth = clonedReq.auth,
-        authInterface = createAuthInterface(auth),
+        authState = {}, // setting this as empty because dry run has not state
```
coditva, line 113:
```suggestion
        authState = {}, // setting this as empty because dry run has no state
```

lib/authorizer/ntlm.js
coditva, line 152: Why changing the key?
```suggestion
        !auth.getAuthState(STATE) && auth.setAuthState(STATE, STATES.INITIALIZED);
```

lib/authorizer/ntlm.js
```diff
@@ -241,7 +240,7 @@ module.exports = {
 
             // Now create the type 3 message, and add it to the request
             auth.set(NTLM_HEADER, authenticateMessage);
```
coditva, line 242: Why keep `ntlm header` in the auth object, but state separately?

lib/authorizer/ntlm.js
sandeep-singh-pstmn: @coditva This auth object is specific to the auth state only, i.e. if a req is using multiple req for auth and wants to share the state b/w the requests, that time this will be used.
that's the reason not we can't set anything other than it
other auth's don't need an auth state

lib/authorizer/ntlm.js
sandeep-singh-pstmn: Header data can be used in scripts, and it's an auth token, so storing this in the auth object makes sense.

lib/authorizer/ntlm.js
sandeep-singh-pstmn: getAuthState / setAuthState We are using it only for the state, so it doesn't make sense to store it inside the `state` key 
the final populated object looks like this
```
{
     NTLM: 'INITIALIZED'
}
```

coditva: But this is the state, no? Why not store in `state` itself? Naming it like `ntlm` feels weird if more state properties are added.

lib/authorizer/ntlm.js
coditva: It feels like we're not sure of what data goes in `auth` object what data goes in `state`. If we're saying that only user data is in `auth` object, then this data does make sense in `state`.

lib/authorizer/auth-interface.js
```diff
@@ -109,6 +110,30 @@ createAuthInterface = function (auth, protocolProfileBehavior, options) {
...
+        /**
+         * @param {String} state - name of state
+         * @return {Object|undefined}
+         */
+        getState: function (state) {
+            if (state && authState && authState[state]) {
+                return authState[state];
+            }
+
+            return undefined;
```
coditva, line 124: This line is redundant.

test/integration/auth-methods/ntlm.test.js
```diff
@@ -466,4 +466,132 @@ var expect = require('chai').expect,
...
+            expect(response1, 'iteration 1, request 2').to.have.property('code', 401);
+            expect(response2, 'iteration 1, request 3').to.have.property('code', 200);
+            expect(response5, 'iteration 2, request 3').to.have.property('code', 200);
+            expect(response8, 'iteration 3, request 3').to.have.property('code', 200);
+            expect(response11, 'iteration 4, request 3').to.have.property('code', 200);
+            expect(response14, 'iteration 5, request 3').to.have.property('code', 200);
+        });
+    });
+
+    describe('with NTLM auth set at collection level with 4 collections', function () {
```
coditva, line 539:
```suggestion
    describe('with NTLM auth set at collection level with 4 requests', function () {
```

lib/authorizer/index.js
```diff
@@ -110,7 +110,8 @@ authorizeRequest = function (request, done) {
 
     var clonedReq = new sdk.Request(request.toJSON()),
         auth = clonedReq.auth,
-        authInterface = createAuthInterface(auth),
+        authState = {}, // setting this as empty because dry run has no state
+        authInterface = createAuthInterface(auth, authState),
```
coditva, line 114: We also need to handle state in `digest` auth.

coditva (commented):
Note: This has breaking changes:
1. NTLM header and state will not be available in `auth` object (in pre-request script).
2. NTLM header cannot be set via the pre-request script.

lib/authorizer/index.js
sandeep-singh-pstmn: On. digest we are setting the user data
- nonce
- realm

coditva (approved):

PR #1399, opened 2024-04-01
Add max body support in edge grid auth  

sandeep-singh-pstmn:
We have added support for the maximum body size in EdgeGrid authentication. As per Akamai's documentation, whenever the body size exceeds the max-body limit, we should trim the body to the maximum allowed size when creating the signature.


lib/authorizer/edgegrid.js
```diff
@@ -129,9 +151,11 @@ var _ = require('lodash'),
 
             return originalReadStream.cloneReadStream(function (err, clonedStream) {
                 if (err) { return callback(); }
+                let remainingMaxBody = maxBody;
```
codenirvana, line 154:
```suggestion
                let readBytes = 0;
```

codenirvana (changes requested):

codenirvana (approved):

PR #1483, opened 2024-11-20
Defer loading the 'serialised-error' module

RaisinTen:
The 'serialised-error' module is only required when there is an error, so this change improves the loading time of this module by loading it only when it is needed for error processing.

Refs: https://github.com/postman-eng/postman-app/pull/16666
cc @postmanlabs/starship

### DIFF
--- a/lib/runner/extensions/event.command.js
+++ b/lib/runner/extensions/event.command.js
@@ -47,6 +47,7 @@ postProcessContext = function (execution, failures) { // function determines whe
         return undefined;
     }
     const serialisedError = require('serialised-error');
+
     return serialisedError(error, true);
 };
 
@@ -475,8 +476,8 @@ module.exports = {
                                     // instance once it is fully supported
                                     result && { cookies: result.cookies });
                             }).catch(function (err) {
-                                const serialisedError = require('serialised-error');
-                                const error = serialisedError(err);
+                                const serialisedError = require('serialised-error'),
+                                    error = serialisedError(err);
 
                                 delete error.stack; // remove stack to avoid leaking runtime internals
                                 this.host.dispatch(EXECUTION_RESPONSE_EVENT_BASE + id, requestId, error);
@@ -528,6 +529,7 @@ module.exports = {
                             // electron IPC does not bubble errors to the browser process, so we serialize it here.
                             if (err) {
                                 const serialisedError = require('serialised-error');
+
                                 err = serialisedError(err, true);
                             }
 
--- a/lib/runner/extensions/event.command.js
+++ b/lib/runner/extensions/event.command.js
@@ -5,7 +5,6 @@ var _ = require('lodash'),
     util = require('../util'),
     sdk = require('postman-collection'),
     sandbox = require('postman-sandbox'),
-    serialisedError = require('serialised-error'),
     ToughCookie = require('@postman/tough-cookie').Cookie,
 
     createItemContext = require('../create-item-context'),
@@ -44,7 +43,11 @@ postProcessContext = function (execution, failures) { // function determines whe
         error.name = ASSERTION_FAILURE;
     }
 
-    return error ? serialisedError(error, true) : undefined;
+    if (!error) {
+        return undefined;
+    }
+    const serialisedError = require('serialised-error');
+    return serialisedError(error, true);
 };
 
 /**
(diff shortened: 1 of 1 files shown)


lib/runner/extensions/event.command.js
```diff
@@ -44,7 +43,12 @@ postProcessContext = function (execution, failures) { // function determines whe
         error.name = ASSERTION_FAILURE;
     }
 
-    return error ? serialisedError(error, true) : undefined;
+    if (!error) {
+        return undefined;
```
tony-go, line 47: @RaisinTen Maybe an opportunity to improve testing :)

RaisinTen:
Hey @appurva21!

> Could you share some details on how much benefit we are getting from this change, specifically on the "load time" of this module?

This shaves ~5ms on my Intel macOS, so the total load time of the module has gone down from ~315ms to ~310ms on my device.

> On a related note, how are we deciding which dependencies to lazy load?

We are trying to improve the main process part of the Postman app launch performance. When the app starts, the `main.js` file runs a bunch of `require()`s which takes up most of the time. We are trying to get rid of all the requires that are not critical for app launch, primarily by deferring the requires until those are needed. Here is the complete list of modules we are focusing on removing from the critical path in this quarter:  https://postmanlabs.atlassian.net/browse/DKTPFN-455. These are arranged in descending order of load time (more details in https://postmanlabs.atlassian.net/wiki/spaces/DKTPFN/pages/5569183798/Main+Process+App+Launch+Targeting+2000ms+p90).

> Lastly, postman-runtime (and all the related request execution logic) is being moved to a separate worker in Postman App soon. Do you think this change still makes sense? I am just trying to understand how are we balancing code readability/simplicity vs. performance gains.

Since 'serialised-error' is only needed during error-cases and it is not necessary for the 'postman-request' module to be usable, not loading this module upfront makes sense to me. In the future, when the worker in postman app would load the 'postman-runtime' module, it's gonna take lesser time if this change is included.

appurva21:
@RaisinTen Thanks. 

> When the app starts, the main.js file runs a bunch of require()s which takes up most of the time. We are trying to get rid of all the requires that are not critical for app launch, primarily by deferring the requires until those are needed

I believe loading postman-runtime is already [deferred](https://github.com/postman-eng/postman-app/pull/13382) when the App starts. AFAIK it only gets loaded on first use (which is after the HTTP request tab loads). How is this change helping with the App start time? I just want to ensure we are reading the metrics right and there is no missing piece of information here.

> Since 'serialised-error' is only needed during error-cases and it is not necessary for the 'postman-request' module to be usable, not loading this module upfront makes sense to me. In the future, when the worker in postman app would load the 'postman-runtime' module, it's gonna take lesser time if this change is included

Agreed. But if this change immediately or in the future, is not helping with the App start time, we would like to take a more holistic approach and audit all such "requires" instead of making a one-of change with no overall strategy (strictly speaking from postman-runtime's point of view)

RaisinTen:
@appurva21 

> I believe loading postman-runtime is already [deferred](https://github.com/postman-eng/postman-app/pull/13382) when the App starts. AFAIK it only gets loaded on first use (which is after the HTTP request tab loads). How is this change helping with the App start time? I just want to ensure we are reading the metrics right and there is no missing piece of information here.

Do you mean that this chunk of `require()`s that run in https://github.com/postman-eng/postman-app/blob/f9fd70698696a392f03651046145a2fc2cf12075/src/main/main.js#L151

Screenshot 2024-11-21 at 4 28 22 PM

is not a part of the `App Relaunch Timings` metric that we measure on the New Relic dashboards? (it measures the time from the app start till the `Send` button is visible - https://github.com/postman-eng/postman-app/blob/f3d8b8deade78411e08b56b4de6ec492405cd5f3/tests/e2e-tests/test/specs/performance/app-relaunch-for-medium-workspace.spec.js#L103)

coditva:
@RaisinTen Yes, we deferred the loading of IPCAgent (and consequently `postman-runtime` et. al.) till it's first use. Currently, the first-use occurs right after the request tab is loaded (when "preview request" is triggered, which dry-runs the request to generate and how the run time headers that are going to be sent with the request).

We can discuss this further if you think that this is getting loaded in the critical path.

RaisinTen:
I can confirm that the runtime module loading is not a part of the `total-app-launch-time` which is used to calculate the p90 Postman app launch time from user metrics, so this PR is no more necessary. I'll close it. Thanks all!

PR #1502, opened 2025-04-13
Add support for parallel Iterations

KhudaDad414:
#### Description
This PR adds support for parallelising iterations in a run.

#### TODO
- [ ] Add more tests.
- [ ] Improve error handling logic.
- [ ] Review the cursor behaviour.


lib/runner/extensions/waterfall.command.js
```diff
@@ -1,65 +1,7 @@
...
-getIterationData = function (data, iteration) {
-    // if iteration has a corresponding data element use that
-    if (iteration < data.length) {
-        return data[iteration];
-    }
-
-    // otherwise use the last data element
-    return data[data.length - 1];
-};
+    { prepareVaultVariableScope, prepareVariablesScope, prepareLookupHash,
```
shamasis, line 3: We sure it is good to move these kind of work to util?

shamasis: on second thought, it maybe okay

PR #1524, opened 2025-09-05
fix: Multiple digest auth headers not filtered by opted algorithm

deve-sh:
Works to fix https://github.com/postmanlabs/postman-app-support/issues/13756

### Context

Since servers can return multiple values for `www-authenticate` headers to present clients with multiple options, we need to filter the right header based on the `auth.algorithm` setting passed by `postman-runtime`'s consumer.

### DIFF
--- a/lib/authorizer/digest.js
+++ b/lib/authorizer/digest.js
@@ -136,12 +136,17 @@ _extractField = function (string, regexp) {
  * there can be more than one header with the same key. So need to loop over and check each one.
  *
  * @param {VariableList} headers -
+ * @param {String} selectedAlgorithm - The user opted algorithm (MD5, SHA-256 etc)
  * @private
  */
-function _getDigestAuthHeader (headers) {
+function _getDigestAuthHeader (headers, selectedAlgorithm) {
     return headers.find(function (property) {
         return (property.key.toLowerCase() === WWW_AUTHENTICATE) &&
-        (_.startsWith(String(property.value).toLowerCase(), DIGEST_PREFIX.toLowerCase()));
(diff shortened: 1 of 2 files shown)


lib/authorizer/digest.js
```diff
@@ -136,13 +136,33 @@ _extractField = function (string, regexp) {
...
+
+            return headerValue.includes(`algorithm=${selectedAlgorithm.toLowerCase()}`);
+        });
+
+        if (headerWithMatchingOptedAlgorithm) {
+            return headerWithMatchingOptedAlgorithm;
+        }
+    }
+
+    return digestAuthHeaders.length ? digestAuthHeaders[0] : null;
```
appurva21, line 165: Any specific reason to return `null` instead of `undefined` as before?

Also:

```js
let matchingHeader;

if (selectedAlgorithm) {
    const targetAlgorithm = selectedAlgorithm.toLowerCase();

    matchingHeader = digestAuthHeaders.find((header) => {
        const headerValue = String(header.value).toLowerCase();

        if (!headerValue.includes('algorithm=')) {
            // This is an MD5 header. Ref: https://datatracker.ietf.org/doc/html/rfc7616
            return targetAlgorithm === 'md5';
        }

        return headerValue.includes(`algorithm=${targetAlgorithm}`);
    });
}

return matchingHeader || digestAuthHeaders[0];
```

Reason:
1. A single return statement makes it easier to follow (unless there is clear bail-out which is not the case here)
2. Avoided doing `selectedAlgorithm.toLowerCase()` twice

lib/authorizer/digest.js
appurva21, line 1: Can we add tests for this?

appurva21 (dismissed):
Approved. Tests to be added.

PR #1527, opened 2025-09-11
Invoke parent's `request` trigger from nested request

deve-sh:


test/integration/runner-spec/run-collection-request.test.js
```diff
@@ -805,4 +805,84 @@ describe('pm.execution.runRequest handling', function () {
                 });
             });
     });
+
+    it('should log nested request to consumer console', function (done) {
```
appurva21, line 809: "should trigger `request` callback"

Logging up to the consumer is not something the runtime is aware of.

(same for changelog as well)

CHANGELOG.yaml
```diff
@@ -2,6 +2,8 @@ unreleased:
   date: 2025-09-11
   fixed bugs:
     - GH-1526 Fixed a bug where nested requests were not aborting the run on failure in assertions
+  new features:
```
appurva21, line 5: Can be treated as bug fix.

appurva21:
Unrelated but @codenirvana suggested to bump the `maxInvokableNestedRequests` from 5 to 10. Can we update that along with this PR?

test/integration/runner-spec/run-collection-request.test.js
deve-sh: Updated

CHANGELOG.yaml
deve-sh: Updated

PR #1530, opened 2025-09-24
feat: Send nested request cached metadata from item to request resolver

deve-sh:
To facilitate better debugging and consumption of cached requests data, we have added cached nested request metadata as well as the `rootRequestId` to a new second argument passed to the nested request resolver.

> This would mean the semver version bump would have to be bumped up to `7.48.0` as the consumers of runtime would have to update the contract used for `requestResolver`

### DIFF
--- a/lib/runner/nested-request.js
+++ b/lib/runner/nested-request.js
@@ -27,7 +27,7 @@ function runNestedRequest ({ executionId, isExecutionSkipped, vaultSecrets, item
         self.state.nestedRequest = _.defaults(self.state.nestedRequest || {}, {
             isNestedRequest: true,
             rootCursor: cursor,
-            rootItemId: item.id,
+            rootItem: item,
             invocationCount: 0
         });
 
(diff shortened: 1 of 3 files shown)


lib/runner/nested-request.js
```diff
@@ -39,8 +39,23 @@ function runNestedRequest ({ executionId, isExecutionSkipped, vaultSecrets, item
                 ' calls have been reached for this request.'));
         }
 
+        let rootRequestEventsList = self.state.nestedRequest.rootItem.events?.all?.() || [],
+            requestMetadataFromItem = null,
+            additionalInfo = { rootRequestId: self.state.nestedRequest.rootItem.id };
+
+        for (const event of rootRequestEventsList) {
+            if (event.script && event.script.requests && event.script.requests[requestToRunId]) {
```
appurva21, line 47: We are using two different patterns for safe checks in a span of a few lines (?. vs &&). Let's stick to one.

lib/runner/nested-request.js
```diff
@@ -39,8 +39,23 @@ function runNestedRequest ({ executionId, isExecutionSkipped, vaultSecrets, item
...
+            }
+        }
+
+        if (requestMetadataFromItem) {
+            additionalInfo.requestMetadataFromItem = requestMetadataFromItem;
+        }
+
         // Should fetch the request from the consumer of postman-runtime & resolve scripts and variables
-        requestResolver(requestToRunId, function (err, collection) {
+        requestResolver(requestToRunId, additionalInfo, function (err, collection) {
```
appurva21, line 58:
```js
context = { rootItemId, referencedRequestMeta: { workspace, collection, path } }
```
- rootItemId since runtime understands item and not request, so keeping it consistent with that
- if we are using `nestedRequest`, we can call it `nestedRequestMeta` as well

appurva21 (changes requested):

lib/runner/nested-request.js
deve-sh: Updated

PR #1549, opened 2026-06-11
feat: per-partition cookie jar via requester.perPartitionCookieJar

scriptonist:
## Summary

Adds opt-in `requester.perPartitionCookieJar` support so each parallel partition/VU can use its own HTTP cookie jar instead of sharing the runner-level jar.

## Changes

- Adds lazy per-partition cookie jar allocation and reset-on-partition-reuse.
- Routes collection requests, `pm.sendRequest`, auth replays, and `pm.cookies.jar()` through the partition jar when enabled.
- Keeps default behavior unchanged; explicit `requester.cookieJar` still takes precedence.
- Adds focused unit and integration coverage for isolation, precedence, lifecycle, and browser-safe test loading.

## Validation

- `npm run test-lint`
- `npm run test-unit`
- Focused per-partition unit/integration specs
- Codecov patch coverage: 100%

### DIFF
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -3,10 +3,10 @@
   new features:
     - >-
       Add requester.perPartitionCookieJar option — allocates an isolated
-      cookie jar per partition (per-VU) in parallelized runs instead of one
-      shared jar per run. Opt-in; an explicitly supplied requester.cookieJar
-      always wins. Jars are reset on stopSinglePartition under the
-      full-fresh-on-reuse contract.
+      cookie jar per parallel partition (equals per-iteration in
+      customParallelIterations mode, where recycled partitions get a fresh
+      jar) instead of one shared jar per run. Opt-in; an explicitly supplied
+      requester.cookieJar always wins.
 
 7.54.0:
   date: 2026-05-05
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -1,3 +1,13 @@
+7.55.0:
+  date: 2026-06-10
+  new features:
+    - >-
+      Add requester.perPartitionCookieJar option — allocates an isolated
+      cookie jar per partition (per-VU) in parallelized runs instead of one
+      shared jar per run. Opt-in; an explicitly supplied requester.cookieJar
+      always wins. Jars are reset on stopSinglePartition under the
+      full-fresh-on-reuse contract.
+
 7.54.0:
   date: 2026-05-05
   new features:
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -1,5 +1,4 @@
-7.55.0:
-  date: 2026-06-10
+unreleased:
   new features:
     - >-
       Add requester.perPartitionCookieJar option — allocates an isolated
--- a/lib/requester/requester-pool.js
+++ b/lib/requester/requester-pool.js
@@ -69,7 +69,20 @@ RequesterPool = function (options, callback) {
 };
 
 RequesterPool.prototype.create = function (trace, callback) {
-    return Requester.create(trace, this.options, callback);
+    var options = this.options;
+
+    // Per-partition cookie jar override (set by http-request.command via
+    // Run#getCookieJarFor in perPartitionCookieJar mode). Shallow-merge
+    // into a copy — the pool's shared options object must never be
+    // mutated per request. The jar reference is frozen here, at requester
+    // creation: a request in flight when its partition is recycled keeps
+    // writing to the old (dead) jar instead of the recycled partition's
+    // fresh one.
+    if (trace && trace.cookieJar) {
+        options = _.defaults({ cookieJar: trace.cookieJar }, options);
+    }
+
+    return Requester.create(trace, options, callback);
 };
 
 module.exports.RequesterPool = RequesterPool;
--- a/lib/runner/extensions/event.command.js
+++ b/lib/runner/extensions/event.command.js
@@ -341,7 +341,17 @@ module.exports = {
 
                     // in the case of nested requests, we want to make sure any exceptions, consoles etc
                     // are tagged to the parent request's scripts
-                    scriptCursor = _.clone(isNestedRequest ? this.state.nestedRequest.rootCursor : cursor);
+                    scriptCursor = _.clone(isNestedRequest ? this.state.nestedRequest.rootCursor : cursor),
+
+                    // Resolve this execution's cookie jar ONCE, at script
+                    // execution start. In perPartitionCookieJar mode a
+                    // partition recycle (stopSinglePartition → resetCookieJar)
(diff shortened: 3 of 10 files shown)


lib/runner/run.js
```diff
@@ -49,6 +49,21 @@ Run = function PostmanCollectionRun (state, options) { // eslint-disable-line fu
...
+    // (e.g., the desktop app passes its own jar to sync with its cookie
+    // store). See Run#getCookieJarFor for the resolution rule and
+    // lib/requester/requester-pool.js for the shared default jar this
+    // option replaces.
+    this.perPartitionCookieJar = Boolean(_.get(this.options, 'requester.perPartitionCookieJar')) &&
+        !_.get(this.options, 'requester.cookieJar');
+
+    if (_.get(this.options, 'requester.perPartitionCookieJar') && _.get(this.options, 'requester.cookieJar')) {
+        // eslint-disable-next-line no-console
+        console.warn('runtime: requester.perPartitionCookieJar is ignored because requester.cookieJar is set');
```
shamasis: This console will output to stdout instead of triggers.console. Intended?

CHANGELOG.yaml
```diff
@@ -1,3 +1,13 @@
+7.55.0:
+  date: 2026-06-10
+  new features:
+    - >-
+      Add requester.perPartitionCookieJar option — allocates an isolated
+      cookie jar per partition (per-VU) in parallelized runs instead of one
```
shamasis: VU is not a runtime concept. We need to see the PR with that vocabulary check in mind.

shamasis (dismissed):
LGTM. The general direction looks okay. I checked to see what is the verdict on memory impact and what has been the teardown strategy for these cookiejars.

The response was. @scriptonist maybe this is of interest.
In default chunked parallel mode (maxConcurrency > 1), a partition runs a contiguous block of N iterations (cyclesPerPartition) and the jar persists across all of them — it's never reset (reset only happens in customParallelIterations recycling). So the unit of isolation is "parallel worker," which equals one iteration only in customParallelIterations mode. 

The CHANGELOG's "isolated cookie jar per partition (per-VU)" is misleading for chunked mode, and "VU" (virtual user) is a load-testing term that appears nowhere else in runtime. Recommend describing it precisely (e.g., "per parallel partition; equals per-iteration in customParallelIterations mode").

CHANGELOG.yaml
shamasis: I also think this is not truly lazy. it creates the cookiejars on start. the ideal way would have been to create cookiejar on first setter/getter access of the cookiejar... it would have avoided the cookiejar class memory footrpint unless a url really sent cookies. one way is to deal with it at the requester layer itself. where the cookiejar pool request is bubbled up .... but that may be too much of a surgical move.

shamasis:
Later, if we genuinely wants per-iteration isolation (fresh jar every iteration even within a worker — i.e. treat each iteration as a stateless request), that's a distinct feature with the opposite tradeoff, and it should be its own explicitly-named option rather than overloading this one. I wouldn't fold it in here; I'd just make sure the docs don't imply this flag does it.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "postman-runtime",
-  "version": "7.54.0",
+  "version": "7.55.0",
```
deve-sh: This will be bumped up with the release automatically.

lib/requester/requester-pool.js
```diff
@@ -69,7 +69,20 @@ RequesterPool = function (options, callback) {
...
+    // Per-partition cookie jar override (set by http-request.command via
+    // Run#getCookieJarFor in perPartitionCookieJar mode). Shallow-merge
+    // into a copy — the pool's shared options object must never be
+    // mutated per request. The jar reference is frozen here, at requester
+    // creation: a request in flight when its partition is recycled keeps
+    // writing to the old (dead) jar instead of the recycled partition's
+    // fresh one.
+    if (trace && trace.cookieJar) {
+        options = _.defaults({ cookieJar: trace.cookieJar }, options);
+    }
```
deve-sh, line 83: Could this be handled in https://github.com/postmanlabs/postman-runtime/pull/1549/changes#diff-1a5f1a12bd740d8bf4926a10904dd06baeda385b44f33659d49a46dcaa9affa2R15 by any chance or does it have to be handled at requester pool creation time?

lib/runner/extensions/event.command.js
```diff
@@ -341,7 +341,17 @@ module.exports = {
...
+
+                    // Resolve this execution's cookie jar ONCE, at script
+                    // execution start. In perPartitionCookieJar mode a
+                    // partition recycle (stopSinglePartition → resetCookieJar)
+                    // swaps the partition's jar reference mid-flight; capturing
+                    // here means a script that was already executing keeps
+                    // writing to the dead jar (harmless, GC'd) instead of
+                    // polluting the recycled partition's fresh jar. Falls back
+                    // to the requester pool's shared jar when not in that mode.
+                    partitionCookieJar = this.getCookieJarFor(cursor);
```
deve-sh, line 354: To use the same cookie jar for nested requests (Via `pm.execution.runRequest`), do we want to use `scriptCursor`?

lib/runner/run.js
scriptonist: Thanks! this was an oversight. 

Fixed in 1525c43e — moved the warning out of the constructor into Run#start and route it through triggers.console, so consumers (desktop app, Newman) surface it instead of it landing on raw stdout. Constructor just records the intent (triggers do not exist yet at construction time).

CHANGELOG.yaml
scriptonist: Corrected the wording in 1525c43e: the jar is allocated on the partition's first request/script (first getCookieJarFor resolution), not on run start and not gated on actual cookie traffic. Truly-lazy allocation on first cookie access would need the requester-layer bubble-up you described — noting it as a possible follow-up, out of scope here.

lib/requester/requester-pool.js
scriptonist: It has to be at create() time. The pool is constructed once per run, but the per-partition jar is resolved per-request via Run#getCookieJarFor and passed in on `trace`. Construction-time `self.options` only knows the run-level/shared jar, so the override can only be applied per requester creation.

lib/runner/extensions/event.command.js
scriptonist: `cursor` already gives nested requests the same jar. The parent script and the nested `pm.sendRequest/runRequest` carry the same `partitionIndex`, and `getCookieJar` For resolves purely on that, so they share one partition `jar. scriptCursor` is the parent-tagged clone used for console/error attribution — resolving on it would not change which jar is picked (same partitionIndex).

package.json
scriptonist: Thanks, reverted

shamasis (approved):

PR #1551, opened 2026-07-01
tests: Stabilize Echo-backed browser and CI tests

deve-sh:
## Summary
- replace browser-sensitive Postman Echo calls with local Echo fixture coverage
- move remaining flaky live Echo and badssl integration assertions onto local fixtures
- skip the known-hanging 308 multipart redirect case, matching the existing 307 skip
- stabilize the synchronous script timeout assertion by waiting for the sandbox timeout payload
- simplify digest fixture URL assertions to use string concatenation instead of constructing URL objects

## Validation
- npm run test-browser
- npm run test-lint
- focused integration checks for inherited auth, timeout, SSL, redirect, history, and v2 regression specs

## Notes
- CodeQL fixture alerts are intentionally left alone because the flagged server code is test-only fixture code.

### DIFF
--- a/test/integration/auth-methods/control-flow/errors.test.js
+++ b/test/integration/auth-methods/control-flow/errors.test.js
@@ -3,23 +3,25 @@ var sinon = require('sinon').createSandbox(),
     AuthLoader = require('../../../../lib/authorizer/index').AuthLoader;
 
 describe('auth control flow', function () {
-    var runOptions = {
-        collection: {
-            item: {
-                name: 'FakeAuth',
-                request: {
-                    url: 'https://postman-echo.com/basic-auth',
-                    auth: {
-                        type: 'fake',
-                        fake: {
-                            username: 'postman',
-                            password: 'password'
+    function getRunOptions () {
+        return {
+            collection: {
+                item: {
+                    name: 'FakeAuth',
+                    request: {
+                        url: global.ECHO_SERVER + '/basic-auth',
+                        auth: {
+                            type: 'fake',
+                            fake: {
+                                username: 'postman',
+                                password: 'password'
+                            }
                         }
                     }
                 }
             }
-        }
-    };
+        };
+    }
 
     after(function () {
         sinon.restore();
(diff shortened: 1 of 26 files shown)


test/integration/sanity/timeouts/sync-scripts.test.js
```diff
@@ -197,13 +197,36 @@ describe('synchronous script timeouts', function () {
...
+                    var prerequest = testrun && testrun.prerequest,
+                        assertions = prerequest && prerequest.firstCall && prerequest.firstCall.args[2];
+
+                    return assertions && assertions[0] && assertions[0].error;
+                }
+
+                function checkTimeoutResult () {
+                    if (!hasTimeoutResult() && Date.now() - startedAt < 3000) {
+                        return setTimeout(checkTimeoutResult, 50);
+                    }
```
deve-sh: Slightly hacky but works

test/integration/inherited-entities/auth.test.js
```diff
@@ -76,7 +76,7 @@ var _ = require('lodash'),
                 response2 = testrun.request.secondCall.args[2];
 
             expect(err).to.be.null;
-            expect(request.url.toString()).to.equal(new URL(global.servers.digest).toString());
+            expect(request.url.toString()).to.equal(global.servers.digest + '/');
```
deve-sh, line 79: While this change does simplify the comparison logic, how does it fix any existing failures because the outcome of both `new URL(...)` and `global.servers.digest + '/'` would be the same right?

Image

deve-sh:
For flakyness verification, re-running all tests to ensure anything didn't slip in

test/integration/auth-methods/control-flow/errors.test.js
```diff
@@ -3,23 +3,25 @@ var sinon = require('sinon').createSandbox(),
...
-            item: {
-                name: 'FakeAuth',
-                request: {
-                    url: 'https://postman-echo.com/basic-auth',
-                    auth: {
-                        type: 'fake',
-                        fake: {
-                            username: 'postman',
-                            password: 'password'
+    function getRunOptions () {
```
deve-sh, line 6: No difference but I don't think a function wrap would be needed for the object here right?

codenirvana (approved):

PR #1553, opened 2026-07-07
Added support for streaming `datasetsResolver` results

souris-ash-postman:
Lets a collection run stream large datasets instead of loading them fully into memory first. Two independent paths, both memory-bounded:

1. Iteration data from a dataset — a run can iterate over a dataset view (one row per iteration). Instead of materialising every row up front, the runner pulls rows lazily, one per iteration.
2. pm.datasets in scripts — when a script calls pm.datasets(...).executeView/Query(...), rows stream to the sandbox in 1000-row batches rather than all at once.


souris-ash-postman:
### 📚 Design & context docs

- **[Streaming datasets through the collection runner: pm.datasets + iteration data](https://postmanlabs.atlassian.net/wiki/x/BoDW6gE)** — the end-to-end spec this PR implements (head-frame + `__pull`/`__cancel` for `pm.datasets`, and the `__streamingIterationData` descriptor).
- **[Streaming big datasets over gRPC/Connect and Arrow](https://postmanlabs.atlassian.net/wiki/x/7YBo6gE)** — the overall data-plane design the runner streaming plugs into.
- **[Iteration-data streaming: 51.2M-row benchmark & memory RCA](https://postmanlabs.atlassian.net/wiki/x/q4Dz6gE)** — bounded-memory evidence and the root-cause analysis behind the `IterationDataStream` sliding window.

codenirvana (dismissed):

codenirvana (approved):

PR #1557, opened 2026-07-31
feat(parallel): per-VU iteration daat for customParallelIterations

saialekhya-001:
## What

Adds a per-VU iteration counter and iteration-data routing for the `customParallelIterations` run mode, so each virtual user (VU) sees its own iteration index and the correct slice of iteration data (`pm.iterationData`) when iterations run in parallel across partitions.

## Changes

- **Per-VU iteration counter** — each partition/VU tracks its own iteration index (`lib/runner/partition-manager.js`, `partition.js`, `run.js`).
- **Parallel command routing** — end-of-partition gating, CR block guards, sentinel cycles, and EOF trigger normalization so rows are dispatched to the right VU (`lib/runner/extensions/parallel.command.js`, `event.command.js`).
- **Full-fresh-on-reuse + late-write drop** — correct state reset on partition reuse and dropping of stale late writes.
- **Settle run-completion callback** in custom mode.
- **Vendored postman-sandbox** — `file:vendor/postman-sandbox-6.7.2-per-vu-variables.tgz` (unreleased build with the per-VU variables + Infinity transform), excluded from publish via `.npmignore`.
- Tests: unit (`custom-parallel-iterations`, `partition-manager`, `per-partition-cookie-jar`) + end-to-end integration (`customParallelIterations.test.js`).


package.json
```diff
@@ -58,7 +58,7 @@
     "performance-now": "2.1.0",
     "postman-collection": "5.3.1",
     "postman-request": "2.88.1-postman.49",
-    "postman-sandbox": "6.7.3",
+    "postman-sandbox": "file:vendor/postman-sandbox-6.7.2-per-vu-variables.tgz",
```
saialekhya-001: [P1] Rebase this sandbox fork onto 6.7.3 before replacing the dependency. The vendored artifact is based on 6.7.2, so this change drops the dataset streaming support already present in 6.7.3. The current required CI confirms the regression: all four dataset pull/stream integration assertions receive zero rows. Carry the sentinel change on top of 6.7.3 (or newer) rather than downgrading the sandbox.

.npmignore
```diff
@@ -51,3 +51,8 @@ docs/
 codecov.yml
 .editorconfig
 .jsdoc-config.json
+
+# - TEMPORARY: exclude the vendored unreleased postman-sandbox build so it never
+#   ships in a published runtime. Remove when postman-sandbox ships the
+#   cycles -1 -> Infinity transform upstream (see vendor/README.md).
+vendor/
```
saialekhya-001: [P1] Do not exclude the artifact that the published package depends on. package.json resolves postman-sandbox from the vendored file, but this rule removes vendor/ from npm pack; consumers of the resulting runtime tarball then fail installation with ENOENT for that path, which is already visible downstream in postman-performance-test#35. Either ship the artifact or depend on a published sandbox version.

lib/runner/partition-manager.js
```diff
@@ -262,6 +270,20 @@ class PartitionManager {
         // Always use iteration 0 since we only have 1 iteration of data.
         // and start from the 0th request position.
         partition.cursor.seek(0, 0);
+
+        // Clear the stop flag before queueing new work — see
+        // updatePartitionVariables for the late-write race this guards.
+        partition.stopped = false;
```
saialekhya-001, line 276: [P1] Keep the stopped generation distinct until its in-flight instruction has completed. stopSinglePartition returns immediately after clearing the queue, so the host can recycle this index and reach this line while the old request or script callback is still pending. Clearing the flag here lets that old callback write variables into the fresh VU, and its partition-processing recursion can also consume the new generation queue. Use a generation token per reuse, or wait for the active instruction to settle before restarting the partition.

PR #1559, opened 2026-08-05
feat(parallel): per-VU iteration data for customParallelIterations

saialekhya-001:
> Supersedes #1557 — that PR was auto-closed when its off-convention head branch `saialekhya/per-vu-parallel-iterations` was renamed to `feat/per-vu-iteration-data`. Same commits, plus the perPartitionCookieJar fix below. Review history is on #1557.

## What

Adds per-VU iteration data + counter for the `customParallelIterations` run mode, so each virtual user (VU) sees its own iteration index and its own row via `pm.iterationData` when iterations run in parallel across partitions. This is the routing the `@postman/performance-test` library depends on for datasets/datafiles as iteration data in parallel perf runs.

## Changes

- **Per-VU iteration counter** — each partition/VU tracks its own iteration index (`partition-manager.js`, `partition.js`, `run.js`).
- **Per-VU iteration data → `pm.iterationData`** — in custom mode the host-injected per-VU row is seeded into BOTH the VU's `_variables` (so `{{key}}` / `pm.variables.get()` and per-partition markers resolve) AND `partition.iterationData`, which `parallel.command` sources the item payload `data` from → `pm.iterationData.get()` works. No sandbox change required.
- **Parallel command routing** — end-of-partition gating, CR block guards, sentinel cycles, EOF trigger normalization.
- **Full-fresh-on-reuse + late-write drop**; settle run-completion callback in custom mode.
- **postman-sandbox** — uses published `6.7.3` (has the `cycles -1 → iterationCount: Infinity` sentinel + `pm.datasets` streaming); the temporary vendored `6.7.2-per-vu` fork was dropped.
- Tests: unit (`custom-parallel-iterations`, `partition-manager`, `per-partition-cookie-jar`) + integration (`customParallelIterations.test.js`, `datasets.test.js`). Full unit + integration suite green locally.

Generated with [Claude Code](https://claude.com/claude-code)


lib/runner/partition.js
```diff
@@ -95,6 +106,16 @@ class Partition {
...
     }
 
+    /**
+     * Re-clones variables from run-level state. Used by stopSinglePartition
+     * to enforce the "full fresh on reuse" contract — when a partition is
+     * recycled by removeUsers/addUsers, the next loop must see a clean
+     * variables scope rather than inheriting the dead VU's mutations.
+     */
+    resetVariables () {
+        this.variables = this._cloneVariables();
```
scriptonist, line 123: `_cloneVariables()` reads `runInstance.state._variables`, but completed parallel scripts write their `_variables` back to that run-global scope in `event.command.js`. This reset therefore clones the previous VU's mutations instead of a pristine baseline (probe: recycled VU saw `fresh:loop-1`). Please reset custom-mode partitions from immutable initial state, or avoid promoting per-VU variables globally.

lib/runner/partition-manager.js
```diff
@@ -262,8 +270,31 @@ class PartitionManager {
         // Always use iteration 0 since we only have 1 iteration of data.
         // and start from the 0th request position.
         partition.cursor.seek(0, 0);
+
+        // Clear the stop flag before queueing new work — see
+        // updatePartitionVariables for the late-write race this guards.
+        partition.stopped = false;
```
scriptonist, line 276: `stopped` cannot reject results from an older VU after this line clears it. On immediate index reuse, an old in-flight completion can arrive with the same `partitionIndex` and overwrite the new occupant. Please tag executions/partitions with a generation and ignore mismatched completions.

lib/runner/partition.js
saialekhya-001: Fixed in e457b1d3. Root cause confirmed: `event.command.js` wrote every completed script's `_variables` back to the run-global `state._variables` unconditionally, and `_cloneVariables()` clones from that scope on reuse — so a recycled VU inherited the previous occupant's mutations.

Fix: skip the run-global write in parallel mode (`result && result._variables && !this.areIterationsParallelized`). The live per-VU scope is still persisted via `updatePartitionVariables`, so `state._variables` stays the pristine baseline new/recycled VUs clone from.

Added an integration regression (`recycled VU does not inherit the previous occupant's pm.variables (no leak)`) — verified it **fails without the fix** (recycled VU saw `loop-1`) and passes with it.

lib/runner/partition-manager.js
saialekhya-001: Fixed in e457b1d3 with generation tagging, as suggested. Each partition now carries a `generation` (Partition#generation) bumped on `stopSinglePartition`; the generation is stamped onto each queued execution (`coords.partitionGeneration` in parallel.command), and `updatePartitionVariables` drops any write whose generation no longer matches the slot's current generation:

```js
if (generation !== undefined && generation !== partition.generation) { return; }
```

So an in-flight completion from a prior occupant of a reused index is ignored instead of overwriting the new VU. Verified the generation actually reaches the guard on real completions (0/1), and added unit coverage for the guard (matching applies / stale drops / undefined back-compat / stopped / missing).

saialekhya-001:
### CI note: `codecov/project/legacy` is pre-existing and non-blocking

All the checks that matter are green — Lint, the full Tests matrix (18/20/latest × ubuntu/win), Browser Tests, **codecov/patch**, **codecov/project**, and the unit + integration project flags.

The only red check is **`codecov/project/legacy` (34.09% vs 35% target)**, which is **not introduced by this PR**:

- It is **already failing on `develop`** — base (`16cdd37`) legacy coverage is **34.33%**, i.e. below the 35% target regardless of this PR.
- This PR's code is well covered: **unit flag `<90.00%>`, integration flag `<78.57%>`** on the diff. The dip is only on the **legacy flag (`<7.14%>`)** because the "legacy" suite is the old callback-style `test-integration-legacy` tests, which don't exercise a new `customParallelIterations` feature (adding legacy-suite tests for a brand-new parallel mode would be contrived).
- It is **not a required/blocking check** — the PR is `mergeable`; `BLOCKED` reflects the pending review, not this status.

Suggested resolution (repo-owner call, out of scope for this PR): either lower the `legacy` target in `codecov.yml` to match actual repo coverage, or backfill legacy-suite coverage as a separate hygiene task.

saialekhya-001:
### Behaviour in `customParallelIterations` mode (perf runs)

Quick reference for how the iteration/data primitives behave under this PR. A perf run isn't "N iterations over a data file" — it's **V virtual users (VUs)**, each a partition, looping for the whole duration; each loop the host hands the VU one data row (via the distribution strategy).

**`pm.info.iteration`** — each VU's **own** loop counter, 0-indexed, climbing for the whole run (VU-A on its 5th loop reads `4` while VU-B on its 2nd reads `1`). Independent per VU, not a shared global count. Resets to `0` when a VU slot is recycled (full-fresh-on-reuse).

**`pm.info.iterationCount`** — **`Infinity`**. There's no fixed count; the run ends by duration. (Internally sent as the `-1` wire sentinel and rendered as `Infinity` by postman-sandbox ≥ 6.7.3, which this PR uses.) Scripts should not gate logic on `iterationCount` in this mode — use `pm.info.iteration` for a per-VU counter.

**`pm.iterationData`** — the VU's **current row** for this loop (the dataset/datafile row picked by the distribution strategy). Read-only. `pm.iterationData.get('col')` returns the row's value. (This is what the PR routes into the item payload's `data` field so the sandbox exposes it.)

**`pm.variables`** — the same row is also seeded into the VU's **Local** scope, so `{{col}}` and `pm.variables.get('col')` resolve it too. `pm.variables.set(...)` writes to **that VU's own** Local scope — isolated per VU (a recycled VU does not inherit a previous occupant's `set()` values). `pm.variables` never mutates `pm.iterationData`.

Resolution precedence for `pm.variables.get()` is unchanged: **Local → Data (iterationData) → Environment → Collection → Globals**.

_Normal (functional) runs are unaffected: `pm.info.iteration` is 0…N-1, `iterationCount` is the real total, and `pm.iterationData` is `data[iteration]` as before._

